PR #5, opened 2014-04-23
DOC: Improve explanation of the `prec` parameter

### LINKED ISSUE #4: Docs for fmprb.h uclear (to me)
I am reading the docs (http://fredrikj.net/arb/fmprb.html):
```
The precision parameter passed to each function roughly indicates the precision to which calculations on the midpoint are carried out (operations on the radius are always done using a fixed, small precision.)
```
Here is an example of such a function that accepts `prec`:
```
void fmprb_div(fmprb_t z, const fmprb_t x, const fmprb_t y, long prec)
```
So the `x` and `y` ball floating point numbers each contain some precision, e.g.
`x = 1 \pm 3` and `y = 2 \pm 0.01`. What `prec` should I, as a user, set to `prec` if I want to calculate `z = x/y`? It says that the result, i.e. `z`, will be rounded to `prec` bits. So if I set `prec` too low, then I'll lose accuracy. What if I set `prec` too high? If I want Arb to do its best and return me the answer as accurate as possible (given the algorithms that it implements), what should I set for `prec`? Will it return an error if `prec` is too high? Or will it just return `z` with lower accuracy if I set `prec` too high?

It would be nice to update the documentation which answers all these questions.

certik:
Fixes #4. This was adapted from the comment made by Fredrik in that issue.

### DIFF
--- a/doc/source/fmprb.rst
+++ b/doc/source/fmprb.rst
@@ -11,10 +11,30 @@ is a ball which contains the result of the (mathematically exact) operation
 applied to any choice of points in the input balls.
 In general, the output ball is not the smallest possible.
 
-The precision parameter passed to each function roughly indicates the
-precision to which calculations on the midpoint are carried out
-(operations on the radius are always done using a fixed, small
-precision.)
+The precision (*prec*) parameter passed to each function roughly indicates the
+precision to which calculations on the midpoint are carried out (operations on
+the radius are always done using a fixed, small precision.) In other words, the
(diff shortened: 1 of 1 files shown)


fredrik-johansson:
I have added a new page to the documentation called "Using ball arithmetic" (using.rst).

It would be great if you could read it and tell me if something still isn't clear.

(Then we can also close this PR.)

fredrik-johansson:
I claim that this problem is resolved, absent any new objections.

PR #42, opened 2015-07-21
fix two compiler warnings

akobel:
(for warnings reported by clang)

PR #43, opened 2015-07-21
fix two compiler warnings

akobel:
(for warnings reported by clang)
What a pity; now I missed #42...

### DIFF
--- a/arb/const_khinchin.c
+++ b/arb/const_khinchin.c
@@ -38,7 +38,7 @@ arb_const_khinchin_eval_param(arb_t s, ulong N, ulong M, long prec)
     arb_init(h);
 
     if (N < 2) abort();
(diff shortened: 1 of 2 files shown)


fredrik-johansson:
The variable ab helps show.what happens in the algorithm.

Does the warning in clang go away if you change "ab = ab;" to "(void) ab;" ?

akobel:
Yes, that will avoid the warning as well. It's a matter of taste; I understand your reason to have it in there, and that's why I only made the `ab = a - aa` a comment and did not remove it altogether.
I'm not sure whether any computation will be spent on setting of `ab` in that case. I know it's not expensive, though. So if you like that better, I'll change it.

fredrik-johansson:
Your code is fine -- I'm happy to merge this.

PR #44, opened 2015-07-23
correct argument order in the documentation of arb_abs...

akobel:
... and change the argument order in the implementation of arb_abs and arb_neg to match the docs.


fredrik-johansson:
Thanks.

PR #45, opened 2015-08-10
Add various set functions

rickyefarr:
A few functions for ease of use.

### DIFF
--- a/arb.h
+++ b/arb.h
@@ -221,10 +221,6 @@ arb_set_si(arb_t x, long y)
     mag_zero(arb_radref(x));
 }
 
-#define arb_init_set_si(x, y)  \
-    arb_init(x);               \
-    arb_set_si(x, y)
-
 ARB_INLINE void
 arb_set_ui(arb_t x, ulong y)
 {
@@ -232,20 +228,12 @@ arb_set_ui(arb_t x, ulong y)
     mag_zero(arb_radref(x));
 }
 
-#define arb_init_set_ui(x, y) \
-    arb_init(x);              \
-    arb_set_ui(x, y)
-
 ARB_INLINE void
 arb_set_d(arb_t x, double y)
 {
     arf_set_d(arb_midref(x), y);
     mag_zero(arb_radref(x));
 }
-
-#define arb_init_set_d(x, y) \
-    arb_init(x);             \
-    arb_set_d(x, y)
 
 ARB_INLINE void
 arb_set_fmpz(arb_t x, const fmpz_t y)
--- a/arb.h
+++ b/arb.h
@@ -229,6 +229,12 @@ arb_set_ui(arb_t x, ulong y)
 }
 
 ARB_INLINE void
+arb_set_d(arb_t x, double y)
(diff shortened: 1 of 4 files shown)


arb.h
```diff
@@ -229,6 +229,12 @@ arb_set_ui(arb_t x, ulong y)
 }
 
 ARB_INLINE void
```
rickyefarr: Didn't know what to do about this line.

arb.h
```diff
@@ -229,6 +229,12 @@ arb_set_ui(arb_t x, ulong y)
...
+arb_set_d(arb_t x, double y)
+{
+    arf_set_d(arb_midref(x), y);
+}
+
+ARB_INLINE void
```
rickyefarr: Wish I could make this line white without adding an extra new line character.  Should I just add the function to the end of the file so that this doesn't happen?

fredrik-johansson:
Thanks for the patch, Rick. A small fix is needed: the method should zero the radius.

rickyefarr:
I thought it did already upon initialization of the variable.

fredrik-johansson:
Yes, but it could have been changed in between.

    arb_init(x);

    arb_const_pi(x, 64);  /* now x has a radius */
    arb_add(y, x, z, 64);

    /* do something else with x */
    arb_set_d(x, 1.5);

rickyefarr:
Ah, I see.  Didn't think about that.  I'll change that momentarily.  Also, I have a bunch of these little functions like this.  I hate to say it, but I may have to add them one at a time until I get used to github.

arb.h
fredrik-johansson: No, the diff is fine. There's problem with it.

fredrik-johansson:
Sure. I take it that you want to add more functions as part of this pull request, before I merge and close it?

rickyefarr:
Yeah, if that's okay with you.  Figured I wouldn't open one up for each little function of the same type.

rickyefarr:
That's all for now.  Eventually, I'd like to add conversions for uintmax and support for longs.

fredrik-johansson:
Could you add these functions to the documentation?

What about uintmax and longs? The ui/si functions are meant to take a 64-bit integer on 64-bit systems. I don't think there's a real need to support 128-bit integers directly.

rickyefarr:
I'm adding them to documentation now.  The only reason I said that was because mpfr has a function for this uintmax.  Also, when I said longs, I meant long double.

rickyefarr:
Wasn't sure what you wanted with the macros in the documentation.  The first section starts out as "Types, macros and constants" but it doesn't seem appropriate for the macros added here would be included in that section.

fredrik-johansson:
Could you just change the macros to functions instead?

fredrik-johansson:
I'm not sure all these combined init_set methods are really necessary though.

rickyefarr:
You're right.  The init_set methods are not really necessary.  I find myself forgetting to use them anyway.  I'll remove and re-submit.

PR #47, opened 2015-08-31
Add _arb_vec_reverse and _acb_vec_reverse

rickyefarr:
Reverse the order of elements within a vector.  

Note that these functions that could be used by arb_poly and acb_poly to find the reciprocal polynomial.


rickyefarr:
Sorry about the messy history.

rickyefarr:
If you decide to add these functions, should I add documentation?  I don't recall much documentation regarding the vector functions.

fredrik-johansson:
There are already _arb_poly_reverse and _acb_poly_reverse methods.

Do you want methods in the _arb_vec namespace too?

PR #49, opened 2015-09-27
Add _arb_vec_pow_ui

rickyefarr:
This is a useful function.  If you agree, @fredrik-johansson, I'll add arb_vec_pow_si and the acb_vec_pow_ui/si as well.  Also, should I add documentation for vec functions?

### DIFF
--- a/arb.h
+++ b/arb.h
@@ -902,6 +902,14 @@ _arb_vec_set_powers(arb_ptr xs, const arb_t x, long len, long prec)
 }
 
 ARB_INLINE void
+_arb_vec_pow_ui(arb_ptr res, arb_srcptr vec, ulong p, long len, long prec)
+{
+    long i;
+    for(i = 0; i < len; i++)
+        arb_pow_ui(res + i, vec + i, p, prec);
+}
+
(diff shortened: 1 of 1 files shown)


fredrik-johansson:
I don't know if every arithmetic operation needs to have a vector version, but feel free to add those functions if you find them useful. I would appreciate documentation!

argriffing:
I tried to fix this by rebasing/merging and adding a docs commit
https://github.com/rickyefarr/arb/commit/72b81cac7af696b5da04f8faa9ca9f0b45af0f0c
https://github.com/rickyefarr/arb/commit/2320b9cada62b42260bbd0ff4a84357085b78dac
but I wasn't pro enough at using github to make it work.

rickyefarr:
@argriffing Thanks for trying to fix it, but if everyone agrees this probably isn't needed. Thoughts?

argriffing:
> Thoughts?

I've written an `arb_vec_extras.c` file for my own use. On one hand I can see how these functions are useful, but on the other hand I can understand the value of not cluttering `arb` with them.

PR #50, opened 2015-10-09
Add arb_mat_eq/ne (and more)!

thofma:
If the functions look fine to you, I can add documentation.


fredrik-johansson:
Look good. Documentation would be nice! You might also copy+replace to add the corresponding functions for acb_mat.

thofma:
If you are ok with the wording, I will add the corresponding acb_mat functions.

fredrik-johansson:
The wording seems ok.

fredrik-johansson:
You added declarations for set/set_round_arb_mat methods. Do you intend to add these functions in this PR?

thofma:
This was actually not intended. But now as it is open and the signatures added, I may as well add these functions here.

PR #52, opened 2015-10-11
using strtod(...) instead atof(...) in arb/test/t-set_str.c to make it MinGW compatible

ahrvoje:
MinGW uses MSVCRT.DLL atof(s) which is not ISO-C compliant. It returns
atof("inf")=0 and the test fails. Seems strtod(s, NULL) works correctly,
with the same intended functionality, and the test passes.

### DIFF
--- a/arb/test/t-set_str.c
+++ b/arb/test/t-set_str.c
@@ -184,7 +184,7 @@ int main()
 
         error = arb_set_str(t, testdata_floats[i], 53);
 
-        x = atof(testdata_floats[i]);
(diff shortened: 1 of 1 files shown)


ahrvoje:
The proposed pull-req is a workaround for MinGW-W64 compiler (as used in Qt) I used.

To make it work for MinGW (MinGW.org project as they refer to themselves) one has to use **_(double) strtold(s, NULL)_**. This variation works in both cases.

fredrik-johansson:
Seems like an OK fix. Thanks!

PR #55, opened 2015-11-11
Fix FLINT include paths

jdemeyer:
Instead of building with a custom include path, `arb` should just have the correct path to the FLINT includes in the sources. This also means that *users* of `arb` do not need to change the include path to build against `arb`.


fredrik-johansson:
Please see issue 24 for discussion of this.

fredrik-johansson:
I've switched from ``flint.h`` to ``flint/flint.h`` in the git master, with a configure hack to create a symbolic link which still allows working against an uninstalled flint. Please test.

BTW, if someone wants to install headers to ``include/arb/arb.h`` instead of ``include/arb.h``, I'll accept a patch that allows doing so *optionally* for now.

PR #62, opened 2015-11-23
ENH: arb_mat_sqr

argriffing:
Save a few multiplications when squaring a matrix. Best for high precision 2x2 matrices. Maybe slightly slower for low-precision large-dimension matrices (i.e. the case where you would want to use BLAS instead).

sqr/mul timing ratios:

|  | n=2 | n=4 | n=10 | n=100 | 
|---|---|---|----|-----| 
| prec=32 | 0.730 | 0.883 | 0.985 | 1.008 | 
| prec=64 | 0.655 | 0.814 | 0.953 | 1.001 | 
| prec=1024 | 0.688 | 0.771 | 0.916 | 0.993 | 
| prec=4096 | 0.555 | 0.738 | 0.871 | 0.986 |

### DIFF
--- a/acb_mat.h
+++ b/acb_mat.h
@@ -141,6 +141,8 @@ void acb_mat_sub(acb_mat_t res, const acb_mat_t mat1, const acb_mat_t mat2, slon
 
 void acb_mat_mul(acb_mat_t res, const acb_mat_t mat1, const acb_mat_t mat2, slong prec);
 
+void acb_mat_sqr(acb_mat_t res, const acb_mat_t mat, slong prec);
+
 void acb_mat_pow_ui(acb_mat_t B, const acb_mat_t A, ulong exp, slong prec);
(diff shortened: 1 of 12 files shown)


fredrik-johansson:
Very nice! Could you make these enhancements for acb_mat too?

You can put your own name in the copyright line if you like. I will, in any case, put your name in the contributor list in the documentation before releasing the next version.

BLAS is obviously very fast, but tricky to use correctly. An easier first step would be to wrap fmpz_mat_mul for multiplying large matrices (which also has the benefit of supporting higher precisions).

With anything other than classical multiplication, the question is what tradeoff you want to make between speed and tightness of error bounds, in particular when you have matrices where the entries vary in size by many orders of magnitude.

argriffing:
Added `acb_mat_sqr`.

PR #64, opened 2015-11-25
TST: acb mat mul test enhancement

argriffing:
Improve `acb_mat_mul` test coverage. I've left the convenience functions in `acb_mat/test/t-mul.c` instead of cluttering the namespace. They could be moved later, or not.

### DIFF
--- a/acb_mat/test/t-mul.c
+++ b/acb_mat/test/t-mul.c
@@ -26,6 +26,59 @@
 #include "acb_mat.h"
 
 
+ACB_INLINE int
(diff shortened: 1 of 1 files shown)


fredrik-johansson:
I think you can drop the number of iterations by a factor 10 to speed up the test.

It's not really optimal to test multiplication this way -- a better way would be to generate random acb_mat:s and check A(B+C) = AB + AC. But your patch is fine of course.

argriffing:
Thanks, I will close this PR and plan to add one checking those kinds of algebraic properties.

PR #67, opened 2015-11-28
ENH: add trace

argriffing:
Add `arb_mat_trace` and `acb_mat_trace`. Much of the code is duplicated.


fredrik-johansson:
Looks perfect.

PR #74, opened 2015-12-10
DOC: arb_poly_set_si does not limit precision

argriffing:


argriffing:
Wait, now I see that line of documentation was for the whole block of preceding functions, the rest of which do have a `prec` argument.

Issue #70, opened 2015-12-01
relative exponential functions (aka exprel aka phi functions)?


argriffing:
There is a family of [relative exponential functions](https://www.gnu.org/software/gsl/manual/html_node/Relative-Exponential-Functions.html#Relative-Exponential-Functions) in gsl. I wonder if it's worth adding some of these to arb? I think the real part of exprel(i*x) is `sinc(x)`, and sinc is listed in https://github.com/fredrik-johansson/arb/issues/35.

index | name | hypgeom | elementary
---|---|---|---
0 | exp(z) | 1F1(1, 1, z) | exp(z)
1 | exprel(z) | 1F1(1; 2; z) | (exp(z) - 1)/z
2 | exprel_2(z) | 1F1(1; 3; z) | 2*(exp(z) - z - 1) / z^2
n | exprel_n(n, z) | 1F1(1; n+1; z) | (n!/z^n) * (exp(x) - sum_k=0^(n-1) x^k/k!)

They are also related to exponential integrators where they are called phi functions, for example in
http://core.ac.uk/download/pdf/1633681.pdf
http://www1.maths.leeds.ac.uk/~jitse/phikrylov.pdf

A recent preprint using exprel:
http://arxiv.org/pdf/1504.01804.pdf

fredrik-johansson:
Well, 1F1 should already do the job. We could perhaps optimize 1F1 for this important special case.

argriffing:
Yes, I don't have a good feel for how good the arb hypgeom functions are compared to their specializations. For example as I was learning arb I looked at a couple naive ways to compute sinc:
1) sin(x) / x
2) 0f1(3/2, -(x/2)^2)
When x is near zero sin(x)/x is not so great, presumably because it doesn't know that the two instances of x are 'the same'. But the hypergeometric formula is good near zero. On the other hand at large x like 50 +/- 1e-16 the hypergeometric formula gives a useless radius like 100 or 1000 whereas the sin(x)/x radius is good. Presumably this is because sin(x) is super optimized and at large x it doesn't matter so much whether the two instances of x in sin(x)/x are the same or not. Anyway, my point isn't that 0f1 is bad, but rather that I don't have good intuition about when a hypgeom function should do the job without extra attention.

fredrik-johansson:
Your observations are completely correct.

[...]

PR #75, opened 2015-12-12
ENH: add sinc

argriffing:
Here's an implementation of `sinc` based on my interpretation of the comments in https://github.com/fredrik-johansson/arb/issues/70#issuecomment-161121373.

It picks among four different evaluations of `sinc` depending on the midpoint and radius of x.

Here's a diagram showing that different methods work better in different regions (midpoint increases top to bottom, radius increases left to right).

im-m6

### DIFF
--- a/arb.h
+++ b/arb.h
@@ -602,6 +602,7 @@ void _arb_sin_cos_pi_fmpq_algebraic(arb_t s, arb_t c, ulong p, ulong q, slong pr
 void arb_sin_cos_pi_fmpq(arb_t s, arb_t c, const fmpq_t x, slong prec);
 void arb_sin_pi_fmpq(arb_t s, const fmpq_t x, slong prec);
 void arb_cos_pi_fmpq(arb_t c, const fmpq_t x, slong prec);
+void arb_sinc(arb_t z, const arb_t x, slong prec);
 void arb_sinh(arb_t z, const arb_t x, slong prec);
 void arb_cosh(arb_t z, const arb_t x, slong prec);
 void arb_sinh_cosh(arb_t s, arb_t c, const arb_t x, slong prec);
--- a/arb/sinc.c
+++ b/arb/sinc.c
@@ -1,0 +1,163 @@
+/*=============================================================================
+
+    This file is part of ARB.
+
+    ARB is free software; you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation; either version 2 of the License, or
+    (at your option) any later version.
+
+    ARB is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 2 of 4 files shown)


fredrik-johansson:
I don't think the 0F1 method should be used. Let's keep this simple and only using elementary functions.

fredrik-johansson:
Also, make sure aliasing works (and test this).

fredrik-johansson:
* Instead of the first arb_overlaps test, you could use arb_get_mag_lower
* You could use a couple bits of extra precision when computing sin(x), to get a slightly better error bound
* You could use min(|x|,1)/2 (with arb_get_mag for |x|) for a tighter derivative bound when x is small

argriffing:
Thanks, I'll try redoing this without 0F1, with aliasing, and with more cleverness about the derivative bound for small x (I understand how this will be needed when 0F1 is no longer used), and with more working precision for sin(x).

argriffing:
> min(|x|,1)/2

Would min(|x|/3, 1/2) be better or is it not cool to use 3 because it's not a power of 2? (second derivative of sinc at 0 is -1/3).

argriffing:
I've made those changes in the latest commit.

fredrik-johansson:
Just one minor thing... you can eliminate the temporary arb_t and arb_set by doing the radius computation before evaluating at the midpoint. In fact, since you have allocated two mag_t variables at the start of the function which are no longer being used, you could recycle them for this purpose (and get rid of the variable 'd' as well).

For the test code, I think you could just test sinc(x) * x = sin(x) instead of comparing with gamma (this should make the test run faster).

argriffing:
I've made those changes.

fredrik-johansson:
Looks good! acb_sinc would be nice to have, too.

argriffing:
> 'acb_sinc'

I see that `acb_sin` is implemented like `sin(x+y*i) = sin(x)*cosh(y) + i*sinh(y)*cos(x)` but I'm not sure what is the analogous identity for `sinc`.

argriffing:
Now I see that the identity is sinc(x+i*y) = (1/(x^2+y^2)) * ( (y*cos(x)*sinh(y) + x*sin(x)*cosh(y)) + i * (x*cos(x)*sinh(y) - y*sin(x)cosh(y)) ) but I don't know if implementing that directly would be any good.

fredrik-johansson:
I think it should be done the same way as arb_sinc: evaluate sin(z)/z, and use a bound for the derivative when near 0.

argriffing:
Great! I see you've implemented `acb_sinc` with the derivative bound `exp(|im(z)|)`.

fredrik-johansson:
Yep. It doesn't always give finite output when it ought to (e.g. for `z = 10 + [10 +/- 10]*I`), but this is a limitation of `acb_div`.

As the next step, it might be useful to have some function to compute the first `n` derivatives of the sinc function, or to compute spherical Bessel functions of the `n` first orders.

PR #76, opened 2015-12-26
MAINT: explicit cast to avoid comparing signed vs. unsigned

argriffing:
This is a small nitpick to get rid of a `-Wsign-compare` compiler warning. I'm not completely sure it's correct, but I think it is.

```
/usr/local/include/flint/flint.h:237:42: warning: signed and unsigned type in co
nditional expression [-Wsign-compare]
 #define FLINT_MIN(x, y) ((x) > (y) ? (y) : (x))
                                          ^
../arb/arb_poly.h:663:12: note: in expansion of macro ‘FLINT_MIN’
     return FLINT_MIN(lo, trunc);
            ^
```


fredrik-johansson:
Thanks. Since lo is <= WORD_MAX and it is implicit that 0 <= trunc <= WORD_MAX, it's correct either way.

PR #83, opened 2016-01-28
ENH: improved structured matrix exp bounds

argriffing:
This PR improves `exp` bounds for matrices that have a structure that is unaffected by taking matrix powers. Examples include diagonal, triangular, or block matrices, possibly under permutation.  This improvement is achieved by detecting matrix entries that are structurally determined and not adding Taylor series truncation errors to these entries. There shouldn't be too much overhead for matrices without this structure.

For structured matrices the `exp` calculation not only has improved bounds, but is also computed faster. The reason for the speedup is that intermediate matrices in the squaring step of 'scaling and squaring' used to have a nontrivial error radius for every matrix entry, whereas after this PR some of these entries have zero error radius.

No new tests have been added, and it's not implemented for `acb` matrices. The existing random matrix fuzz tester for `exp` already generates matrices that have enough structure to provide some test coverage of this PR.

### DIFF
--- a/acb_mat/exp.c
+++ b/acb_mat/exp.c
@@ -27,6 +27,74 @@
 #include "acb_mat.h"
 
 slong _arb_mat_exp_choose_N(const mag_t norm, slong prec);
+void _fmpz_mat_transitive_closure(fmpz_mat_t A);
+
+int
+_acb_mat_is_diagonal(const acb_mat_t A)
+{
+    slong i, j;
+    for (i = 0; i < acb_mat_nrows(A); i++)
+        for (j = 0; j < acb_mat_ncols(A); j++)
+            if (i != j && !acb_is_zero(acb_mat_entry(A, i, j)))
+                return 0;
+    return 1;
+}
+
+int
+_acb_mat_any_is_zero(const acb_mat_t A)
+{
+    slong i, j;
+    for (i = 0; i < acb_mat_nrows(A); i++)
+        for (j = 0; j < acb_mat_ncols(A); j++)
+            if (acb_is_zero(acb_mat_entry(A, i, j)))
+                return 1;
+    return 0;
+}
+
+void
+_acb_mat_exp_get_structure(fmpz_mat_t C, const acb_mat_t A)
+{
+    slong i, j, dim;
+
+    dim = acb_mat_nrows(A);
+    fmpz_mat_zero(C);
+    for (i = 0; i < dim; i++)
+    {
+        for (j = 0; j < dim; j++)
+        {
+            if (!acb_is_zero(acb_mat_entry(A, i, j)))
+            {
+                fmpz_one(fmpz_mat_entry(C, i, j));
+            }
+        }
+    }
+    _fmpz_mat_transitive_closure(C);
+}
+
+void
+_acb_mat_exp_set_structure(acb_mat_t B, const fmpz_mat_t C)
+{
+    slong i, j, dim;
+
+    dim = acb_mat_nrows(B);
+    for (i = 0; i < dim; i++)
+    {
+        for (j = 0; j < dim; j++)
+        {
+            if (fmpz_is_zero(fmpz_mat_entry(C, i, j)))
+            {
+                if (i == j)
+                {
+                    acb_one(acb_mat_entry(B, i, j));
+                }
+                else
+                {
+                    acb_zero(acb_mat_entry(B, i, j));
+                }
+            }
+        }
+    }
+}
 
 /* evaluates the truncated Taylor series (assumes no aliasing) */
 void
@@ -143,6 +211,20 @@ acb_mat_exp(acb_mat_t B, const acb_mat_t A, slong prec)
         return;
     }
 
+    /* todo: generalize to (possibly permuted) block diagonal structure */
+    if (_acb_mat_is_diagonal(A))
+    {
+        if (B != A)
+        {
+            acb_mat_zero(B);
+        }
+        for (i = 0; i < dim; i++)
+        {
+            acb_exp(acb_mat_entry(B, i, i), acb_mat_entry(A, i, i), prec);
+        }
+        return;
+    }
+
     is_real = acb_mat_is_real(A);
 
     wp = prec + 3 * FLINT_BIT_COUNT(prec);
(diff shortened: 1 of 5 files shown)


fredrik-johansson:
Nice. I wanted something like this but didn't know about the term "transitive closure".

Comments:

* Could you add a line to the documentation explaining that this is done?
* For aliasing of B and A to work in arb_mat_exp, A must not be read from after B has been written to.
* Have you made sure that the test code produces adequate coverage (i.e. computing the transitive closure incorrectly results in tests failures)?
* I suppose a public function that outputs the transitive closure (perhaps as an fmpz_mat) could be useful for future use.
* Would it be useful to try to get any information from the largest power A^m computed explicitly when evaluating the Taylor series? For example, what if A is nilpotent?
* This should be done for acb_mat_exp too. I guess you could even figure out which entries must have a zero real/imaginary parts with some more work, but that's probably overkill.
* We may as well exponentiate diagonal matrices directly...

argriffing:
> Could you add a line to the documentation explaining that this is done?

Added `Truncation error is not added to entries whose values are determined by the sparsity structure of A.`

> For aliasing of B and A to work in arb_mat_exp, A must not be read from after B has been written to.

The transitive closure is now computed before writing to B.

> Have you made sure that the test code produces adequate coverage (i.e. computing the transitive closure incorrectly results in tests failures)?

Yes this was both accidentally and deliberately tested :)  But no unit tests have been added specifically for the implementation of Warshall's algorithm.

> I suppose a public function that outputs the transitive closure (perhaps as an fmpz_mat) could be useful for future use.

I'm not sure where this would go...

> Would it be useful to try to get any information from the largest power A^m computed explicitly when evaluating the Taylor series? For example, what if A is nilpotent?

Yes I've thought about nilpotent matrices, but I'd prefer not to be that clever in this PR.

> This should be done for acb_mat_exp too. I guess you could even figure out which entries must have a zero real/imaginary parts with some more work, but that's probably overkill.

I'll leave this until the `arb_mat` modifications are suitable for merging, because the `acb_mat` implementation will probably involve copying code which I'd rather do only once.

> We may as well exponentiate diagonal matrices directly...

I've added this, and the test matrix sampler has been modified to generate diagonal and triangular matrices with greater probability than before.

fredrik-johansson:
OK, looks good to me.

argriffing:
I've copied this from `arb` to `acb`.

argriffing:
> Would it be useful to try to get any information from the largest power A^m computed explicitly

While it's on my mind, I thought I'd jot down these examples...
 - If A^m is the zero matrix then `A` is nilpotent so you don't need any truncation error
 - If A^m has no entry that is zero, then you don't have to look at the matrix structure
But intermediate examples can be misleading if you are not careful..
 - If A is like  {{0, 1, 0}, {0, 0, 1}, {1, 0, 0}} then A^3 is the identity matrix
 - If A is like {{-1, 0}, {0, -1}} then the first order truncated Taylor approximation of exp(A) is the zero matrix

argriffing:
> Would it be useful to try to get any information from the largest power A^m computed explicitly when evaluating the Taylor series? For example, what if A is nilpotent?

I see how to do this, in particular you can get a max truncation power that is more informative than 'zero' or 'infinity' for each entry of A. It's a bit more complicated than computing just the transitive closure, but I think it's still O(N^3). Maybe there is a simpler way, but here is what I was thinking:
 1. Use something like Tarjan's algorithm to get the 'condensation' DAG of strongly connected components (SCCs). This also indexes the SCCs in reverse topo sort order, and it gives the partition by providing a map from vertex index to SCC index.
 2. Compute the transitive closure of the condensation.
 3. For each SCC pair (s, t), use the transitive closure to determine whether any walk from s to t in the condensation includes a cycle-containing SCC. An SCC contains a cycle if its size is > 1 or if any of its vertices has a loop.
 4. For each SCC pair (s, t), determine the max length path from s to t in the condensation. You can do this efficiently because the condensation is a DAG.

This wouldn't be too difficult to add, but I'm not sure about the code organization or the interface. It has some graph theoretic stuff that seems to be a bit outside the scope of arb. Would it make sense to depend on a graph library? Or to add a separate arb module for graph stuff? Or to keep the code within 'arb_mat/exp.c', or even to just not include this functionality because it is too complicated?

fredrik-johansson:
I wouldn't depend on a graph library just for this. If it's no more than a few hundred lines of code, can be tested, gives a marked improvement for realistic input (examples would be nice), and doesn't slow down the general case too much, I think it could be implemented as is. I have nothing against putting a handful of non-numerical utility fmpz_mat functions in the arb_mat module.

argriffing:
Here's something that seems to work, but it's not a real PR. It feels like there is too much code in that exp.c file.
https://github.com/argriffing/arb/tree/scc-for-matrix-power-series

fredrik-johansson:
I need to look at it in detail, but I think it will not be so bad if you move the connectivity stuff to a separate file.

There should be a high level mathematical description for _arb_mat_exp_sufficient_N and _arb_mat_exp_get_structure and these could perhaps be named differently.

Could you simplify things a bit by using an fmpz_mat for _arb_mat_exp_sufficient_N? You use inf as a special value, but this could presumably be -1, or perhaps even some convenient finite positive value larger than the potential number of terms.

argriffing:
> Could you simplify things a bit by using an fmpz_mat for _arb_mat_exp_sufficient_N? You use inf as a special value, but this could presumably be -1

Yes, I see how this would be an improvement. I'll plan to submit a more modest PR first.

argriffing:
> There should be a high level mathematical description

I don't know the math jargon, but I want to call it something like an 'entrywise nilpotence degree' of a non-negative matrix, although this may not quite work because it's not really an entrywise function.

The nilpotence degree of a matrix is given at https://en.wikipedia.org/wiki/Nilpotent_matrix, and I think it is equivalent to an unnecessarily complicated definition like "the smallest non-negative `k` so that `A^N = 0 when N >= k`." This high level mathematical description is convenient because it has an entrywise generalization: "the smallest non-negative `k_{ij} so that (A^N)_{ij} = 0 when N >= k_{ij}`." I'd like to add an efficient implementation of this function to fmpz_mat_extras and restricted to non-negative `A`, but I still don't know what an actual mathematician would call it...

fredrik-johansson:
Sure. I don't mean that any fancy terms have to be used, there should just be a clear statement of what it computes.  So if I understood correctly, this computes an upper bound for the nilpotence degree by just looking at the sparsity structure. It doesn't tell you anything about a full nilpotent matrix like [[1,-1],[1,-1]]. Maybe "structural nilpotence degree bound" is a good term.

argriffing:
Yes, it would be exact for matrices with non-negative matrices. In fact I was thinking of only allowing non-negative matrix inputs to the function. This would obviate the need for qualifiers like 'structural' or 'bound'. For matrices with negative entries like `[[1,-1],[1,-1]]` or `[[5, -3, 2], [15, -9, 6], [10, -6, 4]]` you can get upper bounds on the (entrywise) nilpotence degree bounds by passing the entrywise absolute value matrix. Just to be clear, this function would be fmpz_mat->fmpz_mat -- it is not simply computing a single degree or degree bound for the entire matrix.

PR #84, opened 2016-01-28
arb_get_ubound/lbound_arf

thofma:
I have added `arb_get_ubound/lbound_arf` functions. So far there were only functions for absolute values for elements of type `arb`. I was in need of them in julia/Nemo, but thought they could be useful also in arb itself.


fredrik-johansson:
The rounding modes should be ARF_RND_FLOOR / ARF_RND_CEIL.

thofma:
Fixed it.

fredrik-johansson:
Thanks!

PR #93, opened 2016-02-23
MAINT: document sparsity-related arb_mat functions and move them to a separate file

argriffing:
This PR only adds docs and reorganizes code, and it should not affect the behavior of existing public functions.

It adds the new file `arb_mat/adjacency.c` containing `_arb_mat_entrywise_nonzero_round_up` and `_arb_mat_count_nonzero_round_up` which are wrapped in the header file by `arb_mat_adjacency` and `arb_mat_count_nonzero`. This breaks the existing `arb_mat` convention of 1:1 correspondence between public function names and .c file names, so if that convention is important then this PR should not be merged as-is.

This PR has not included analogous changes to `acb_mat`.

### DIFF
--- a/arb_mat.h
+++ b/arb_mat.h
@@ -301,21 +301,20 @@ void arb_mat_trace(arb_t trace, const arb_mat_t mat, slong prec);
 
 /* Sparsity structure */
 
-void _arb_mat_entrywise_nonzero_round_up(fmpz_mat_t A, const arb_mat_t src);
+void arb_mat_entrywise_is_zero(fmpz_mat_t dest, const arb_mat_t src);
 
-ARB_MAT_INLINE void
-arb_mat_adjacency(fmpz_mat_t A, const arb_mat_t src)
-{
-    _arb_mat_entrywise_nonzero_round_up(A, src);
-}
+void arb_mat_entrywise_not_is_zero(fmpz_mat_t dest, const arb_mat_t src);
 
-slong _arb_mat_count_nonzero_round_up(const arb_mat_t src);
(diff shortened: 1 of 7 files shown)


fredrik-johansson:
I don't see the need to break convention here, nor for the separate "underscore" methods.

The documentation needs to be clear about whether these methods are counting entries that are certainly nonzero or entries that are not certainly zero. Ideally, the function names should also reflect this.

arb_mat_entrywise_nonzero_round_up and arb_mat_count_nonzero_round_up are descriptive names but slightly obscure.

One possible naming convention would be:

arb_mat_entrywise_not_is_zero (!arb_is_zero entrywise)

arb_mat_entrywise_is_zero (arb_is_zero entrywise)

arb_mat_entrywise_is_nonzero (arb_is_nonzero entrywise)

etc.

It's a bit redundant to have both the operation and the logical negation, but there's no convenient way to "not" an fmpz_mat.

Something like "get_mask" could be an alternative to "entrywise"

Correspondingly, you could have arb_mat_count_not_is_zero, arb_mat_count_is_zero, etc. Though here the negations are certainly redundant since you can just subtract the count from the matrix size.

argriffing:
If I use the function naming conventions that you've suggested and I keep the .c filename vs. function name convention, then it is OK to add multiple new similar .c files?

fredrik-johansson:
Sure. As long as the functions are useful. If we end up with dozens of such functions, we could perhaps put them together.

argriffing:
updated function names and documentation

fredrik-johansson:
It looks all right.

Issue #69, opened 2015-11-30
hadamard product?


argriffing:
I'm not sure if the [Hadamard product](https://en.wikipedia.org/wiki/Hadamard_product_(matrices)) would be in scope or how the function would be named. This is just elementwise multiplication, and it's a different concept than [Hadamard matrices](https://en.wikipedia.org/wiki/Hadamard_matrix) or the [Hadamard inequality](https://en.wikipedia.org/wiki/Hadamard%27s_inequality) which are already in flint/arb.

fredrik-johansson:
It's a rather trivial operation, and relatively uncommon. Is it worth the trouble? I guess it could be handy sometimes when you're (ab)using matrices as 2D arrays. I'm not going to object to a patch.

I think arb_mat_mul_entrywise would be a good name. In fact, the analogous function for polynomials would be handy, too; Hadamard product of power series is a fairly important operation.

argriffing:
> I guess it could be handy sometimes when you're (ab)using matrices as 2D arrays.

This is totally what I was going to do. Maybe using arb through `nemo` would be more appropriate.

fredrik-johansson:
There's nothing wrong with making the C interface convenient though, if you want to use it for that :-)

PR #94, opened 2016-02-23
ENH: mul_entrywise

argriffing:
Includes docs and tests. Closes https://github.com/fredrik-johansson/arb/issues/69. Test execution times are under control because test settings are copied from matrix multiplication, and matrix products are slower than entrywise products.


fredrik-johansson:
Thanks!

PR #95, opened 2016-02-24
MAINT: move transitive_closure to a new fmpz_mat_extras module

argriffing:
Includes a test.


fredrik-johansson:
OK, this seems like a reasonable way to do it. Documentation is missing, but this could be added when there is more than one method in this module.

PR #96, opened 2016-02-24
[dirichlet] add conrey type to handle logs + char type

pascalmolin:
NOT TESTED, for the moment it just compiles, I know some errors (e.g. the FIXME in group_init.c : the generators have to be lifted mod q)
this commit is just a proof of concept, I can make it work very soon (at least take care of the test file).

Contains:

- try to handle even and odd components the same way in the dirichlet group

- switch from phi_q_odd to smaller expo = exponent of the group (all character orders divide this number, and a character of that order exists)

- use conrey logarithm to reuse log and to loop efficiently over the group

  (see the diff on l.c, only 1 log in computation instead of 2 * q)

- I have defined a char object but not used it in l.c

### DIFF
--- a/acb_dirichlet/conrey_log.c
+++ b/acb_dirichlet/conrey_log.c
@@ -1,0 +1,91 @@
+/*=============================================================================
+
+    This file is part of ARB.
+
+    ARB is free software; you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation; either version 2 of the License, or
+    (at your option) any later version.
+
+    ARB is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License
+    along with ARB; if not, write to the Free Software
+    Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA
+
+=============================================================================*/
+/******************************************************************************
+
+    Copyright (C) 2015 Jonathan Bober
+    Copyright (C) 2016 Fredrik Johansson
+
+******************************************************************************/
+
+#include "acb_dirichlet.h"
+
+/* todo: modular arithmetic
+   discrete log can be computed along exponents or using p-adic log
+*/
+
+void
+acb_conrey_init(acb_conrey_t x, const acb_dirichlet_group_t G) {
+    x->log = flint_malloc(G->num * sizeof(ulong));
+}
+
+void
+acb_conrey_one(acb_conrey_t x, const acb_dirichlet_group_t G) {
+    ulong k;
+    for (k = 0; k < G->num ; k++)
+      x->log[k] = 0;
+    x->n = 1;
+}
+
+void
+acb_conrey_clear(acb_conrey_t x) {
+    flint_free(x->log);
+}
+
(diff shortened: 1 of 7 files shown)


fredrik-johansson:
I have some reservations about names, and what appears to be some duplication. We shouldn't have both acb_dirichlet_chi and acb_dirichlet_char_eval; if an acb_dirichlet_char_t is introduced, then this should be the canonical way to specify a character.

Indentation should be 4 spaces everywhere.

acb_dirichlet/conrey_log.c
```diff
@@ -0,0 +1,91 @@
...
+{
+    ulong k, pk, gk;
+    /* even part */
+    if (G->neven >= 1)
+      x->log[0] = (m % 4 == 3);
+    if (G->neven == 2)
+    {
+        ulong q_even = G->q_even;
+        ulong g2 = 5;
+        ulong m2 = (m % 4 == 3) ? n_negmod(m, q_even) : m % q_even;
```
fredrik-johansson: This is broken: n_negmod assumes that the input is already reduced.

fredrik-johansson:
A question: what would be the most convenient way to evaluate a character at the vector of primes 2 <= p <= bound?

pascalmolin:
As for duplication: in the first place I had everything with the character type (which performs some reductions useful when the character is used many times). I dupplicated to allow long entries in response to Jonathan's objection, which I quite understand. There is a decision to be made here, I have no strong opinion on that point.

Anyway I consider your feedback as an encouragement and I am going to go further (and follow your coding guidelines).

As for the evaluation at primes: I guess that this concerns the more general question of computing the Dirichlet series for some length. I have three algorithms, the best depends on the parameters (size of the group and its factorization compared to the length of the coefficients).

In this draft I have written only the simplest and most efficient, when one needs more than a large proportion of the group size.

I have another algo which does the same at each prime (already written, just need to translate).

Finally when the order of the group is really large and does not factor, the best is to compute the log of each prime and propagate by sieving (corresponds to a Euler product). I have written this but without an important improvement: in this case, if one needs k logarithms,
each baby-step giant-step table for prime p should be reused and its size increased to size ~sqrt(pk) instead of sqrt(p) [for ulong sizes I would guess it is not worth considering better than bsgs].

fredrik-johansson:
Right, I had in mind computing the truncated Dirichlet series. Already with the Hurwitz zeta algorithm, instead of computing each Hurwitz zeta series separately, one should be able to compute the single truncated Dirichlet L-series, and only compute the Euler-Maclaurin formula tails separately.

For the Riemann zeta function, Arb only evaluates 1/n^s directly when n is prime; when n is composite, 1/n^s is obtained by a single multiplication. The same trick should clearly be used for Dirichlet L-series.

About the rest -- mainly, I think it needs to be really obvious how to use the interface without knowing about Conrey logarithms. So it should be obvious how to create Dirichlet characters, iterate over them, iterate over their values, and compute their values at single points... and none of this should be slower than necessary.

Extra "middle steps" can indeed be added where justified for performance, but it is at least good to try to keep any extra functions as orthogonal as possible.

pascalmolin:
OK. Then I go on with a character type (we can revert to log when necessary
if there is a good reason to do so), and add what I have. This should be
done by the end of this week (no more time right now that's why I urged to
submit something incomplete).

As for Dirichlet series, I have some thoughts on the subject based on my
experience on L functions via Fourier inversion, we can discuss that
afterwards.

Pascal

2016-02-24 15:27 GMT+01:00 Fredrik Johansson <notifications@github.com>:

PR #97, opened 2016-02-25
ENH: use sparsity structure more intelligently for matrix power series

argriffing:
Avoid adding unnecessary matrix power truncation terms to functions of nilpotent or "partially nilpotent" matrices. An example (maybe unnecessarily contrived) is the nilpotent matrix
```
A =
[0 2^-100 0]
[0 0 2^100]
[0 0 0]
```
for which the matrix exponential with `prec=32` used to have large error bounds
```
[1 +/- 0, 7.88860905221012e-31 +/- 2.0842e+13, 0.5 +/- 1.321e+43]
[0 +/- 0, 1 +/- 0, 1.26765060022823e+30 +/- 2.0842e+13]
[0 +/- 0, 0 +/- 0, 1 +/- 0]
```
but which is now recognized as having no error:
```
[1 +/- 0, 7.88860905221012e-31 +/- 0, 0.5 +/- 0]
[0 +/- 0, 1 +/- 0, 1.26765060022823e+30 +/- 0]
[0 +/- 0, 0 +/- 0, 1 +/- 0]
```

The PR includes a test, but no new docs yet. It also includes some new utility functions in `fmpz_mat_extras`.

### DIFF
--- a/acb_mat.h
+++ b/acb_mat.h
@@ -94,6 +94,18 @@ void acb_mat_set_arb_mat(acb_mat_t dest, const arb_mat_t src);
 
 void acb_mat_set_round_arb_mat(acb_mat_t dest, const arb_mat_t src, slong prec);
 
+ACB_MAT_INLINE int
+acb_mat_is_empty(const acb_mat_t mat)
+{
+    return (mat->r == 0) || (mat->c == 0);
+}
+
+ACB_MAT_INLINE int
+acb_mat_is_square(const acb_mat_t mat)
+{
+    return (mat->r == mat->c);
+}
+
 /* Random generation */
 
 void acb_mat_randtest(acb_mat_t mat, flint_rand_t state, slong prec, slong mag_bits);
--- a/acb_mat.h
+++ b/acb_mat.h
@@ -333,6 +333,22 @@ void acb_mat_charpoly(acb_poly_t cp, const acb_mat_t mat, slong prec);
 
 void acb_mat_trace(acb_t trace, const acb_mat_t mat, slong prec);
 
+/* Sparsity structure */
+
+void acb_mat_entrywise_is_zero(fmpz_mat_t dest, const acb_mat_t src);
+
+void acb_mat_entrywise_not_is_zero(fmpz_mat_t dest, const acb_mat_t src);
+
+slong acb_mat_count_is_zero(const acb_mat_t mat);
+
+ARB_MAT_INLINE slong
+acb_mat_count_not_is_zero(const acb_mat_t mat)
+{
+    slong size;
+    size = acb_mat_nrows(mat) * acb_mat_ncols(mat);
+    return size - acb_mat_count_is_zero(mat);
+}
+
 #ifdef __cplusplus
 }
 #endif
--- a/acb_mat/count_is_zero.c
+++ b/acb_mat/count_is_zero.c
@@ -1,0 +1,45 @@
+/*=============================================================================
+
+    This file is part of ARB.
+
+    ARB is free software; you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation; either version 2 of the License, or
+    (at your option) any later version.
+
+    ARB is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License
+    along with FLINT; if not, write to the Free Software
+    Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301 USA
+
+=============================================================================*/
+/******************************************************************************
+
+    Copyright (C) 2016 Arb authors
+
+******************************************************************************/
+
+#include "acb_mat.h"
+
+slong
+acb_mat_count_is_zero(const acb_mat_t mat)
+{
+    slong nz;
+    slong i, j;
+    nz = 0;
+    for (i = 0; i < acb_mat_nrows(mat); i++)
+    {
+        for (j = 0; j < acb_mat_ncols(mat); j++)
+        {
+            if (acb_is_zero(acb_mat_entry(mat, i, j)))
+            {
+                nz++;
+            }
+        }
+    }
+    return nz;
+}
--- a/acb_mat/entrywise_is_zero.c
+++ b/acb_mat/entrywise_is_zero.c
@@ -1,0 +1,43 @@
+/*=============================================================================
+
+    This file is part of ARB.
+
+    ARB is free software; you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation; either version 2 of the License, or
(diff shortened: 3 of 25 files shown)


fredrik-johansson:
One question I have is: what is the worst case complexity of this method? Have you done any benchmarking?

I'm thinking that a fast way to compute the sparsity structure of A^N for nonnegative A would be to use matrices that have boolean entries subject to the "arithmetic operations"

a + b = a OR b
a * b = a AND b

Now you can just compute A^N using O(n^3) matrix multiplication x O(log N) binary powering. This should be quite fast if you do it naively, but if you bit pack the matrices and do 64 entries at a time in the matrix multiplication, it should be extremely fast.

argriffing:
> One question I have is: what is the worst case complexity of this method?

It's proportional to n^3, where n is the order of the square matrix (the number of rows and columns). It can't be faster than transitive closure which is not faster than matrix multiplication. Tarjan's algorithm is O(|V| + |E|) which I guess for `fmpz_mat` is like n^2. The other ingredients downstream of Tarjan's algorithm are not difficult to analyze for complexity, and the most complex have three nested loops so they are n^3.

> Have you done any benchmarking?

No, not at all.

> I'm thinking that a fast way to compute the sparsity structure of A^N for nonnegative A would be to use matrices that have boolean entries subject to the "arithmetic operations"
>
> a + b = a OR b
> a * b = a AND b

Yes I'd spent some time not finding this in FLINT. It's not GF(2) and it's not 2-adic. I think the jargon is 'boolean semiring' (http://marcpouly.ch/pdf/internal_100712.pdf) which supports matrices, vectors, polynomials, etc.

> Now you can just compute A^N using O(n^3) matrix multiplication x O(log N) binary powering. [if you hack bits] it should be extremely fast.

Right. I agree that it would be more efficient to use actual bits instead of arbitrary precision integer types to represent zeros and ones. I also agree that it may be better to use an algorithm with slightly worse computational complexity if the code is less complicated or if it can be implemented particularly efficiently so that it is faster for practical inputs.

Just to be clear, the fmpz matrix computed by the new method does not give the sparsity structure of A^N. I guess that you know this, but I'm not completely sure from the comments. Also, an entry of the sparsity structure of A^N is not enough to determine whether truncation error should be added to that entry of the truncated series. For example if the original matrix has the sparsity structure of a directed cycle graph, then (A^N)_{ij} will probably be zero even though (A^M)_{ij} will be nonzero for some M greater than N.

fredrik-johansson:
> No, not at all.

OK, I would like to see some ballpark figure before merging this, just to be sure.

> Just to be clear, the fmpz matrix computed by the new method does not give the sparsity structure of A^N.

Yes, that's just for detecting if it's actually nilpotent.

Otherwise, if you are happy with the code, I can merge it, but documentation would be nice.

argriffing:
> I would like to see some ballpark figure

I reran an old analysis that included TIMEIT_START/TIMEIT_STOP for `arb_mat_exp` of all combinations of {dense, almost dense, lower triangular, strict lower triangular, diagonal} x {n=2, n=4, n=10} x {prec=32, prec=64, prec=1024} x {entries with zero midpoint have nonzero error, entries with zero midpoint have zero error} with `fmpq_mat_randtest`->`arb_mat` matrices and I see the following summary using callgrind:

| Function | Incl. | Called |
| --- | --- | ---: |
| arb_mat_sqr_classical | 0.9252 | 176474 |
| _arb_mat_exp_taylor | 0.0545 | 18941 |
| fmpz_mat_entrywise_nilpotence_degree | 0.0024 | 11848 |

so for this benchmark the squaring was like 20x as slow as the truncated power series which was like 20x as slow as the truncated power series. I don't know how this should be extrapolated.

I suppose a separate code branch could be added for matrices that are actually nilpotent (no entry of the entrywise_nilpotence_degree matrix is -1). This could avoid the scaling and squaring, and use enough terms in the power series to avoid truncation. I don't know enough about 'rectangular evaluation' of power series to know if this would be a good idea, but I see that in any case N should be truncated according to the nilpotence degree if a matrix is actually nilpotent.

> documentation would be nice.

I'll plan to add docs to this PR before merging.

argriffing:
Added some docs, and implemented the newer exp tricks for `acb_mat`.

fredrik-johansson:
"If the matrix *B* computed by this function contains any negative value, then this means that *A* is nilpotent."

This sounds wrong...

fredrik-johansson:
Also, this is trivial nitpicking, but did you think of another name instead of entrywise_nilpotence_degree? entrywise does suggest a matrix operation really done entrywise without regard to the structure.

argriffing:
> This sounds wrong...

Thanks, I've just fixed that.

> Also, this is trivial nitpicking, but did you think of another name instead of entrywise_nilpotence_degree? entrywise does suggest a matrix operation really done entrywise without regard to the structure.

No I can't think of a better name and I agree it's not good. Thanks for bringing this up, I don't think it's such trivial nitpicking; in my opinion, much of the value of this kind of project is in the notation it provides.  Maybe Bill Hart would have an opinion on the interface/names for these functions?

fredrik-johansson:
I think the test code for exp needs to be stronger. Right now removing "|| fmpz_cmp_si(fmpz_mat_entry(S, i, j), N) > 0" doesn't cause any failure.

argriffing:
> I think the test code for exp needs to be stronger. Right now removing "|| fmpz_cmp_si(fmpz_mat_entry(S, i, j), N) > 0" doesn't cause any failure.

I've increased test coverage to fix this issue. Previously the exp test matrices included dense, diagonal, and upper/lower triangular matrices. Now strict triangular matrices (which are nilpotent) are included in the tests.

There is no hurry to merge, I'll try to address the function name issue with a later commit.

argriffing:
renamed the function and updated docs

argriffing:
This PR is getting out of control, I'll close it and add another small one like https://github.com/fredrik-johansson/arb/pull/95 without the strongly-connected-components stuff.

argriffing:
> matrices that have boolean entries subject to the "arithmetic operations"
>
> a + b = a OR b
> a * b = a AND b

I'm planning to add this as `bool_mat`, implemented with an array of `int` (int appears to be used for booleans in flint and arb) rather than hacking bits or abusing fmpz_mat. The purpose would not be to improve efficiency, but to more clearly describe certain functions related to paths and closures. For example the 'reflexive transitive closure' AKA 'all pairs reachability' of `A` would be `sum_{k=0}^\infty A^k` in this semiring. Ultimately the idea is to improve bounds for arb/acb matrix power series.

fredrik-johansson:
If you prefer. I think `fmpz_mat` is still fine for this purpose, and `bool_mat` without bit hacking is not going to be a whole lot faster.

It's standard to use `int` for boolean values in C, but `char` is just as idiomatic if you need arrays of them, assuming that you don't want to go through the trouble of hacking bits.

argriffing:
> you can just compute A^N using O(n^3) matrix multiplication x O(log N) binary powering.

For a non-negative matrix if you only want to detect nilpotency or get the nilpotency degree then I think you can get this in O(n^2). But entrywise series truncation is O(n^3). I have the O(n^2) detection and degrees here https://github.com/argriffing/arb/tree/bool-mat, but it's not a real PR yet.

PR #100, opened 2016-02-29
ENH: bool_mat

argriffing:
Proposing `bool_mat` (matrix over boolean or/and semiring) with tests and docs. This PR does not modify existing code to use the proposed module.


fredrik-johansson:
Looks OK.

argriffing:
The `bool_mat_entry` could possibly be split into a getter and setter. This would facilitate switching to an eventual packed-bits implementation, but disadvantages include "YAGNI" and divergence from other flint/arb modules.

fredrik-johansson:
It should probably be a getter and a setter. People find weird uses for available code all the time; might as well leave the door open for using a packed representation.

PR #103, opened 2016-03-02
MAINT: bool_mat entry getter and setter

argriffing:
addresses https://github.com/fredrik-johansson/arb/pull/100#issuecomment-190920000

Uses `bool_mat_get_entry` and `bool_mat_set_entry` instead of `bool_mat_entry`.  Sometimes I'm unsure how to write notation in the docs, for example 'sets the entry to x' vs. 'sets the value of the entry to x', or if 'entry' and 'element' mean different things.

### DIFF
--- a/bool_mat.h
+++ b/bool_mat.h
@@ -51,14 +51,19 @@ bool_mat_struct;
 
 typedef bool_mat_struct bool_mat_t[1];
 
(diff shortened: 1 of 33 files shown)


fredrik-johansson:
entry, element, tomayto, tomahto

PR #104, opened 2016-03-02
MAINT: fix mem leak in bool_mat complement involution test

argriffing:
I wasn't clearing a bool_mat in one of the unit tests.

The valgrind tests could be streamlined to check a whole module at once for error/leaks. For this PR I am doing this by running `$ make valgrind MOD=bool_mat` and then inspecting `cat build/bool_mat/test/t-*.valgrind` visually for anything that looks like an error or leak, but it could be automated. It would also be nice if the number of iterations were reduced by maybe 10x when running valgrind, but I know that arb doesn't have that iteration control yet.

### DIFF
--- a/bool_mat/test/t-complement.c
+++ b/bool_mat/test/t-complement.c
@@ -84,18 +84,17 @@ int main()
     for (iter = 0; iter < 10000; iter++)
     {
         slong m, n;
-        bool_mat_t A, C, CC;
+        bool_mat_t A, C;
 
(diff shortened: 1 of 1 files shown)


fredrik-johansson:
What you suggest could be useful, though generally, I run valgrind only for the affected test when I change one function/test substantially, so it's not a big problem to invoke it manually and wait a few minutes for it to finish.

PR #105, opened 2016-03-02
MAINT: use bool_mat for structure detection in matrix exp

argriffing:
This code churn uses `bool_mat` for the 'structural' parts of matrix exp. It shouldn't change the behavior. In particular, it is not yet trying to be clever about nilpotent matrices or about non-nilpotent matrices for which some entries have only a finite number of nonzero terms in the matrix power series.


fredrik-johansson:
Nothing to complain about.

PR #110, opened 2016-03-08
enh: linear algebra functions for symmetric positive definite (spd) matrices

argriffing:
This is based on Cholesky factorization, and includes docs and tests.

As an example of the accuracy improvement, here are error magnitudes for inverses and determinants of various sizes of Pascal matrices with prec=96, using the generic `inv` and `det` vs. the `spd_inv` and `spd_det` (check mark indicates the solution is exact, and &#x2717; indicates utter failure).

| n | inv | spd_inv | det | spd_det |
| --- | --- | --- | --- | --- | --- |
| 2 | &#x2713; | &#x2713; | &#x2713; | &#x2713; |
| 3 | &#x2713; | &#x2713; | &#x2713; | &#x2713; |
| 4 | 2.198e-25 | &#x2713; | 3.33e-27 | &#x2713; |
| 5 | &#x2713; | &#x2713; | &#x2713; | &#x2713; |
| 6 | 1.02e-19 | &#x2713; | 4.597e-24 | &#x2713; |
| 7 | 1.267e-16 | &#x2713; | 1.888e-22 | &#x2713; |
| 8 | 2.271e-13 | &#x2713; | 1.235e-20 | &#x2713; |
| 9 | 6.285e-14 | &#x2713; | 9.331e-23 | &#x2713; |
| 10 | 2.288e-06 | &#x2713; | 7.317e-17 | &#x2713; |
| 11 | 0.007739 | &#x2713; | 4.992e-15 | &#x2713; |
| 12 | 39.61 | &#x2713; | 4.873e-13 | &#x2713; |
| 13 | 5.275e+04 | &#x2713; | 1.22e-11 | &#x2713; |
| 14 | 2.064e+09 | &#x2713; | 6.186e-09 | &#x2713; |
| 15 | 1.849e+13 | &#x2713; | 6.991e-07 | &#x2713; |
| 16 | 1.757e+17 | &#x2713; | 9.328e-05 | &#x2713; |
| 17 | 3.521e+18 | &#x2713; | 2.561e-05 | &#x2713; |
| 18 | 1.24e+27 | &#x2713; | 0.9967 | &#x2713; |
| 19 | &#x2717; | 1.706e+30 | 1.883e+04 | 1.486e-07 |
| 20 | &#x2717; | 4.526e+34 | 1.073e+10 | 5.248e-06 |
| 21 | &#x2717; | 2.31e+40 | 3.194e+16 | 0.003213 |
| 22 | &#x2717; | &#x2717; | 7.323e+23 | &#x2717; |
| 23 | &#x2717; | &#x2717; | 1.751e+35 | &#x2717; |


argriffing:
I've cited the method I used for matrix inversion using cholesky decomposition, but I also looked at their BSD-licensed MATLAB code. Maybe that information would have to be included somehow? It should be compatible with ARB's more restrictive license.

fredrik-johansson:
Why not the LDL version, avoiding square roots?

Remark: I guess this can be done for Hermitian acb_mats too?

Remark: I believe that the correct way to avoid exploding intervals in Gaussian elimination (and probably also Cholesky), at least when the matrix is well-conditioned, is to do a floating-point computation and then determine an a posteriori error bound.

fredrik-johansson:
By the way, for a good way to compute the determinant of an arb_mat, write the matrix as (A + E) * 2^n where A is an integer matrix with entries less than 2^prec (or maybe 2^((1+alpha)*prec)) and invoke fmpz_mat_det, then use a bound such as theorem 2.12 in http://www4.ncsu.edu/~ipsen/ps/simax070477.pdf

Incidentally, this ‎will be exact for integer matrices.

argriffing:
> Why not the LDL version, avoiding square roots?

No reason except that functions related to LL were simpler to implement, and both LL and LDL give the 2x speedup vs. LU.  I guess I'll plan to switch to LDL, but I don't want the scope of this PR to start including matrices that are not spd, and in particular I really don't want to mess with LBL where B is block diagonal.

> I guess this can be done for Hermitian acb_mats too?

Yes, I haven't implemented that yet for a few reasons. One is that I personally don't need it yet, another is that it will involve copy-pasting code so I'd like to get the `arb_mat` versions in mergeable condition first. A third, maybe silly, reason is that I'm not sure what to call those matrices in the interface. Maybe `hpd` for Hermitian positive definite (as opposed to `spd` for symmetric positive definite)?

>  I believe that the correct way to avoid exploding intervals [...] is to do a floating-point computation and then determine an a posteriori error bound.

OK I may keep this in mind for future work.

> for a good way to compute the determinant of an arb_mat [...]

I've just deleted the determinant stuff from this PR. I didn't really need it, and it may be distracting. But the method that you've suggested looks interesting.

argriffing:
> both LL and LDL give the 2x speedup vs. LU

Now that I look at this more closely, I think LDL uses like 3/2 as many multiplications as LL, so it can be slower.

fredrik-johansson:
> Now that I look at this more closely, I think LDL uses like 3/2 as many multiplications as LL, so it can be slower.

The question is whether it gives better error bounds.

argriffing:
I'm working on adding it as `arb_mat_ldl` and `arb_mat_solve_ldl_precomp` in analogy to `arb_mat_cho` and `arb_solve_cho_precomp`, and to `arb_mat_lu` and `arb_solve_lu_precomp`.

fredrik-johansson:
> Maybe hpd for Hermitian positive definite (as opposed to spd for symmetric positive definite)?

Sure.

> I've just deleted the determinant stuff from this PR.

I guess it could have some uses... your choice.

fredrik-johansson:
> I'm working on adding it as arb_mat_ldl and arb_mat_solve_ldl_precomp in analogy to arb_mat_cho and arb_solve_cho_precomp, and to arb_mat_lu and arb_solve_lu_precomp.

Do you want to do so in this PR?

argriffing:
> Do you want to do so in this PR?

Yup, I'm adding it here. Maybe also `arb_mat_inv_cho_precomp` and `arb_mat_inv_ldl_precomp` where `arb_mat_spd_inv` is an alias for one of those.

argriffing:
>>Now that I look at this more closely, I think LDL uses like 3/2 as many multiplications as LL
>
>The question is whether it gives better error bounds.

Here's a discussion where the author tried to address this question directly http://stackoverflow.com/questions/23112943/time-complexity-of-cholesky-decomposition-for-the-ldl-form. My feeling is that his answer is wrong there, and that LDL doesn't need the 3/2 slowdown if you factor the terms cleverly, for example using something like the `v` vector in http://www.physics.arizona.edu/~restrepo/475A/Notes/sourcea-/node66.html but I haven't looked into that yet.

argriffing:
I've added LDL, with tests and docs. The Golub and Van Loan text has a detailed algorithm for LDL that is about as fast as LL, so I've used that. For LDL I've also figured out how to adapt Krishnamoorthy's double-speed inv() trick, although unlike the LL case I haven't found any reference code or explicit algorithm. This trick is an alternative to computing an inverse by blindly solving AX=B where B=I. Regarding speed and accuracy, I haven't studied this rigorously. It's not clear to me that LDL is generally better than LL for symmetric positive definite matrices.

argriffing:
Some notes in case they are useful later:
 - Wikipedia's LDL decomposition pseudocode currently uses twice as many multiplications as the Golub and Van Loan algorithm. This has caused confusion on the internets, including on StackOverflow.
 - The LL and LDL decompositions are about the same speed as each other.
 - For both LL and LDL, the special-cased matrix inverse function that depends on the decomposition is about twice as fast as using the decomposition of A to solve AX=B with B set to the identity matrix.
 - In its associated publication, the special-cased matrix inverse function is reported to have slightly better, or at least not worse, accuracy than solving AX=I. This result may or may not extend to interval matrices.

argriffing:
> Remark: I believe that the correct way to avoid exploding intervals in Gaussian elimination (and probably also Cholesky), at least when the matrix is well-conditioned, is to do a floating-point computation and then determine an a posteriori error bound.

Yes I think this would be way better. I noticed a paper that implements this, "Solving and Certifying the Solution of a Linear System" by Nguyen Hong Diep and Nathalie Revol.

fredrik-johansson:
There is also section 6.2 in http://www.texmacs.org/joris/ball/ball.pdf and Rump's paper http://www.ti3.tuhh.de/rump/intlab/ActaNumerica2010.pdf

argriffing:
Algorithm 10.7 in Rump's paper looks pretty good; it's the basis for INTLAB's `verifylss`. An `arb` implementation would benefit from a hypothetical `arf_mat`...

PR #116, opened 2016-03-11
MAINT: 30 -> MAG_BITS

argriffing:
Try to replace some magic numbers with the appropriate name. I'm not sure if this is correct. The motivation was to help look into how `arb_intersection` might be implemented.


fredrik-johansson:
This constant could be anything (2, for example) since the only purpose is to determine the sign of the result, but MAG_BITS is as good as anything else.

Issue #114, opened 2016-03-10
arb_intersection


argriffing:
I see that there is an `arb_union` but not an `arb_intersection`. FWIW I would use this in my own code for the multivariate interval newton method.

fredrik-johansson:
How would you define the intersection of two disjoint balls?

argriffing:
Maybe use an integer return value?

argriffing:
Oh github is adding likes and votes to comments... not sure how I feel about that.  Maybe &#x1f44e;

fredrik-johansson:
Yes, that sounds like a sensible interface.

PR #120, opened 2016-03-14
ENH: arb_intersection

argriffing:
Sets an output value to an interval containing an intersection, and has the same integer return value as `arb_overlaps()`.
Closes https://github.com/fredrik-johansson/arb/issues/114.
This doesn't include an `acb` implementation yet. I guess that should be added after the `arb` version is reviewed and before it is merged.


fredrik-johansson:
Looks fine, though one of the test loops isn't randomizing the precision.

argriffing:
The test coverage has been increased.  I noticed that there is no `acb_union`, so now I'm less eager to implement `acb_intersection` in this PR.

PR #125, opened 2016-04-02
ENH: arb min and max

argriffing:
Includes tests and docs. The implementation and tests follow `arb_union` and the docs follow those of the analogous `arf` functions.


fredrik-johansson:
Looks good.

PR #126, opened 2016-04-05
ENH: frobenius norm

argriffing:
Includes docs, and an identity involving this norm and the matrix trace has been added to the trace tests. It's implemented for real and complex matrices. A variant that computes an `arb_t` interval instead of a `mag_t` upper bound could be useful too, but that is not implemented here.

### DIFF
--- a/acb_mat.h
+++ b/acb_mat.h
@@ -150,6 +150,8 @@ void acb_mat_transpose(acb_mat_t mat1, const acb_mat_t mat2);
 
 void acb_mat_bound_inf_norm(mag_t b, const acb_mat_t A);
 
+void acb_mat_bound_fro_norm(mag_t b, const acb_mat_t A);
+
 /* Arithmetic */
 
 void acb_mat_neg(acb_mat_t dest, const acb_mat_t src);
--- a/acb_mat/bound_fro_norm.c
+++ b/acb_mat/bound_fro_norm.c
@@ -1,0 +1,59 @@
+/*=============================================================================
+
+    This file is part of ARB.
+
+    ARB is free software; you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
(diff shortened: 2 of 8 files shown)


fredrik-johansson:
accepted

argriffing:
Do you have a naming preference for a variant that gives an `arb_t` interval or a `mag_t` lower bound?  The motivation is an algorithm that involves ratios of norms.

fredrik-johansson:
I guess just `arb_mat_XXX_norm` for the first one.

For a lower bound, the word `lower` could be appended, like most `mag` methods. An `upper` could be inserted into the name of the upper bound version too, for symmetry... though it's somewhat unnecessary since norm bounds clearly will be upper bounds if no context is given.

Playing with some naming schemes...

    arb_mat_xxx_norm(arb_t, ...)
    arb_mat_xxx_norm_upper(mag_t, ...)
    arb_mat_xxx_norm_lower(mag_t, ...)

    arb_mat_xxx_norm(arb_t, ...)
    arb_mat_xxx_norm_bound(mag_t, ...)
    arb_mat_xxx_norm_bound_lower(mag_t, ...)

    arb_mat_xxx_norm(arb_t, ...)
    arb_mat_bound_xxx_norm(mag_t, ...)
    arb_mat_bound_xxx_norm_lower(mag_t, ...)

fredrik-johansson:
By the way, "fro" sounds a little weird to me. Instinctively, I'd take "frobenius" or even "frob". But I could get used to it. Besides, it aligns with "inf", "one" and "two", so that's nice.

The next question is whether it should be "xxx_norm" or "norm_xxx". The first reads out like English, the second makes more sense structurally...

argriffing:
> By the way, "fro" sounds a little weird to me.

Yeah I was using the notation from numpy/scipy/matlab/octave. I agree it sounds weird and it could be changed.

PR #127, opened 2016-04-06
add interval frobenius norm

argriffing:
This renames fro->frobenius, and tests have been moved from `t-trace.c` to new `t-frobenius_norm.c` files.

fwiw, the interval implementation was inspired by @jwbober's [arb_mat_L2norm](https://github.com/jwbober/ntlib/blob/arb/src/modform/arb_extras.c#L53).


fredrik-johansson:
Looks good.

PR #128, opened 2016-04-06
symmetric diagonalization

argriffing:
Contributed by @jwbober, do not merge yet...

### DIFF
--- a/arb_mat.h
+++ b/arb_mat.h
@@ -343,6 +343,8 @@ void arb_mat_charpoly(arb_poly_t cp, const arb_mat_t mat, slong prec);
 
 void arb_mat_trace(arb_t trace, const arb_mat_t mat, slong prec);
 
+int arb_mat_symmetric_diagonalization(arb_mat_t D, arb_mat_t P, const arb_mat_t A, slong prec);
+
 /* Sparsity structure */
 
 void arb_mat_entrywise_is_zero(fmpz_mat_t dest, const arb_mat_t src);
--- a/arb_mat.h
+++ b/arb_mat.h
@@ -343,7 +343,14 @@ void arb_mat_charpoly(arb_poly_t cp, const arb_mat_t mat, slong prec);
 
 void arb_mat_trace(arb_t trace, const arb_mat_t mat, slong prec);
 
-int arb_mat_symmetric_diagonalization(arb_mat_t D, arb_mat_t P, const arb_mat_t A, slong prec);
+int _arb_mat_jacobi_diagonalization(arb_mat_t D, arb_mat_t P, const arb_mat_t A, slong prec);
+
+ARB_MAT_INLINE int
+arb_mat_symmetric_diagonalization(
+        arb_mat_t D, arb_mat_t P, const arb_mat_t A, slong prec)
+{
+    return _arb_mat_jacobi_diagonalization(D, P, A, prec);
+}
 
 /* Sparsity structure */
 
@@ -367,4 +374,3 @@ arb_mat_count_not_is_zero(const arb_mat_t mat)
 #endif
 
 #endif
-
--- a/arb_mat/symmetric_diagonalization.c
+++ b/arb_mat/symmetric_diagonalization.c
@@ -1,0 +1,408 @@
+/*=============================================================================
+
+    This file is part of ARB.
+
+    ARB is free software; you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation; either version 2 of the License, or
+    (at your option) any later version.
+
+    ARB is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License
+    along with ARB; if not, write to the Free Software
+    Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA
+
+=============================================================================*/
+/******************************************************************************
+
+    Copyright (C) 2016 Jonathan Bober
+
+******************************************************************************/
+
+#include "arb_mat.h"
+
+void arb_set_exact(arb_t x) {
+    mag_zero(arb_radref(x));
+}
+
+void arb_twobytwo_diag(arb_t u1, arb_t u2, const arb_t a, const arb_t b, const arb_t d, slong prec) {
+    // Compute the orthogonal matrix that diagonalizes
+    //
+    //    A = [a b]
+    //        [b d]
+    //
+    // This matrix will have the form
+    //
+    //    U = [cos x , -sin x]
+    //        [sin x, cos x]
+    //
+    // where the diagonal matrix is U^t A U.
+    // We set u1 = cos x, u2 = -sin x.
+
+    if(arb_contains_zero(b)) {
+        // this is not quite right (doesn't set error intervals)
+        arb_set_ui(u1, 1);
+        arb_set_ui(u2, 0);
+        return;
+    }
+    arb_t x; arb_init(x);
+
+    arb_mul(u1, b, b, prec);            // u1 = b^2
+    arb_sub(u2, a, d, prec);            // u2 = a - d
+    arb_mul_2exp_si(u2, u2, -1);        // u2 = (a - d)/2
+    arb_mul(u2, u2, u2, prec);          // u2 = ( (a - d)/2 )^2
+    arb_add(u1, u1, u2, prec);          // u1 = b^2 + ( (a-d)/2 )^2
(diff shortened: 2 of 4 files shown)


argriffing:
with prec=64 the matrix {{1, 2^60}, {0, 2^60}} causes trouble

fredrik-johansson:
From a quick glance at the code, it looks like it finishes when abs(x1) < 2^(-prec * 0.9). This probably needs to be changed to a relative test, taking into account the magnitude of the other entries in the matrix. There probably needs to be a stopping criterion based on iter too.

argriffing:
This PR might work now, but it needs docs and the code could be cleaned up a bit.

Two tests are included: one tests that the characteristic polynomial of the original matrix contains zero at diagonal entries of the resulting diagonal matrix, and another tests that when the decomposition is multiplied out the resulting interval matrix contains the original matrix.

Two modifications/hacks were required to make the tests pass and the calculation not hang. The first hack sets off-diagonals of the 2x2 diagonalization step to zero by fiat, as part of what I understand to be a heuristic stage of the algorithm which should not affect interval correctness.  The second hack is to normalize columns at the end of the interval calculation stage so that the interval of each column norm includes 1.0.

jwbober:
I'm not going to really look at this right now, but your "first hack"
sounds like a good idea, and something I wish I had thought of trying. The
"second hack" is not really a hack at all. I view it as a judgement call
about whether the function should merely give correct eigenvectors or a
should give a correct orthogonal matrix. The normalization is probably the
right thing to do, since in practice it shouldn't cost much precision for
reasonable matrices.

There's probably some work that can be done in choosing a good internal
working precision and stopping criteria to get best possible results.
What's in there now is just something that seemed to work for me in
practice. It's not something I really gave any thought to.

argriffing:
@jwbober It works for me too, and I think it could be merged unless you guys want to polish it further. Thanks for contributing it!

argriffing:
There is an inconsistency in the return value documentation "If the eigenvalues can be certified as unique, then a nonzero int is returned" vs. the line `if(unique_eigenvalues) return 0;`.

argriffing:
More nitpicky stuff (sorting eigenvalues, choosing eigenvector signs) has been added into this PR. It seems to work great with a DCT-II example; the P and D outputs overlap the entries of the 'true' matrices which can be computed analytically. This kind of test is more stringent than the unit tests included here.

fredrik-johansson:
* The algorithm needs to be documented, in particular how error bounds are certified.
* What happens with matrices that have repeated eigenvalues?
* Some tests with a matrix constructed from known eigenvalues might be a good idea.
* _arf_hypot could use arf_addmul
* ARF_RND_NEAR is used though this isn't supported yet (meaning an arbitrary rounding is done, which is fine here, but technically undefined)
* There should be some stopping criterion based on iter, unless it can be proved that it always stops in a reasonable number of steps.
* What does it do with non-finite input?
* I'm still concerned about the absolute tolerance test. What happens if you feed in a matrix where the entries are scaled by 2^100000 or 2^-100000?
* Instead of stopping at a tolerance of 0.9 * prec, the numerical part could perhaps be done with a working precision of 1.1 * prec for a tolerance of prec.  (However, this would create some slowdown due to crossing the word boundary when the input 64 or 128 bits, so it's maybe not necessary.)
* At high precision, the tolerance is probably too conservative; with a 5x5 matrix and 1000-bit precision, it's unlikely that 100 bits will be lost.

argriffing:
> What happens if you feed in a matrix where the entries are scaled

I've added a test for DCT-II decomposition, and it has problems with scaled matrices which were not apparent for the unscaled test matrices. My understanding of the math is that diagonalizing a scaled matrix should give the same result as scaling the eigenvalues in the diagonalization of the unscaled matrix. But this test is failing.

fredrik-johansson:
As another perhaps even more stringent test, you could perhaps try generating a random vector of rational eigenvalues with varying magnitudes, with some of them very close to each other, apply a random similarity transformation, and check that you can recover the eigenvalues to desired accuracy with a precision-increasing loop.

argriffing:
It seems to do OK with an eigenvalue separation test; I think it's good when the precision becomes high. But I think there's some kind of bug where it will certify incorrect eigenvalue intervals for un-distinguished eigenvalues at low precision for scaled matrices.

For example in the DCT test I think that sometimes none of the computed eigenvalue intervals enclose the true spectral radius at low precision. It knows that it cannot distinguish the eigenvalues at such a low precision and it knows that it cannot give useful bounds for the eigenvector entries, but sometimes it also gets the eigenvalue intervals wrong.

jwbober:
The low precision DCT test puzzled me for a while, but I think that the
certification is working correctly. It just needs to be documented
correctly.

Each of the returned eigenvalue intervals does contain an eigenvalue, which
is all that is checked. When the computed eigenvalues are not unique, the
certification makes no guarantee that all the eigenvalues of the matrix are
contained in the output.

This also makes me realize that without more work and/or thought, _all_ of
the eigenvectors need to have a large error radius when the eigenvalues
aren't unique; if just one pair of eigenvalue intervals overlaps, it may be
the case that those intervals are enclosing a single eigenvalue and another
eigenvalue is completely missed, or some other interval encloses two
eigenvalues, and in that case the certification of the eigenvectors is
wrong, because it assumes a one-to-one correspondence between eigenvalues
and the computed eigenvalue intervals.

This seems like reasonable behavior. A different sort of reasonable
behavior when the eigenvalues are not unique might be to somehow cheaply
compute bounds for the largest and smallest eigenvalues and set all of the
error intervals on the eigenvalues to be large enough to contain both of
them. Then in some sense there would be a one-to-one correspondence between
eigenvalues and the computed intervals, but in another sense the answer
might actually contain less information.

For my own purposes right now, I need the eigenvalues of the matrix I'm
diagonalizing to be unique, and when they aren't I compute a different
matrix to diagonalize. So I don't actually know what I want the function to
do when the eigenvalues aren't unique.

jwbober:
For the behavior when there are repeated eigenvalues (or when the working
precision isn't enough to distinguish them) I thought it might be
reasonable to compare with what arb does for polynomial root finding. Right
now there seems to be only one polynomial root finding function which
doesn't start with an underscore, which is acb_poly_find_roots(). It has
the behavior that when it doesn't find n distinct roots, it returns some
roots, but makes no guarantee that the returned intervals contain all of
the roots.  That's similar to the currently implemented behavior here.

In this case, we do know that all of the eigenvalues are real, so it may be
possible to do a little better. For example, it might be possible give
reasonable output for reasonable input and to guarantee somehow that all of
the eigenvalues are in the output intervals and every output interval which
doesn't overlap with any other output interval contains exactly one
eigenvalue. (This should generically be the case for "reasonable" input,
but perhaps it's not straightforward and/or cheap to prove.)

The error bounds are currently computed as follows:

First, given an (approximate) eigenvalue and eigenvector pair (lambda,v)
for the matrix A, let (A - lambda*I)v = e. Then there is an eigenvalue in
the interval lambda +/- |e|/|v|. So we momentarily treat lambda and v as
exact and compute e in interval arithmetic (including using whatever
intervals the entries of A have) and then set the error on that eigenvalue
to be an upper bound for the absolute value of |e| / |v|.

If we found n non-overlapping intervals (for an n x n matrix) then we've
found all of the eigenvalues, and they are distinct. In this case, for the
same lambda (which is now an interval), v, and e as above, if d is the
minimum possible distance from lambda to another eigenvalue, then the
distance from v to the exact eigenvector corresponding to the unique
eigenvalue in the interval represented by lambda is bounded by |e|/d.

It would be good if someone other than me looked carefully the computation
of the error intervals to make sure that I'm computing what I think I am
computing.

(By the way, Andy Booker suggested these error bounds to me, and so they
are what I used. Similar sorts of bounds are described at
http://people.math.gatech.edu/~klounici6/2605/Lectures%20notes%20Carlen/chap3.pdf,
I think, and those notes could also be useful for someone trying to do a
bit better and/or figure out good working precision and stopping criteria.)

argriffing:
Just as a heads up, I'm not actively working on this PR right now. So if you guys want to close it, or fork it, or make a new one that's OK with me.

fredrik-johansson:
No problem. I don't have time for it either right now, but maybe later. In any case, I will need to be 100% convinced that it's correct before I can merge it.

I'm thinking that the code could be modularized a bit more. How about extracting the verification step to a separate function?

The Jacobi algorithm should work even with repeated eigenvalues, right? I wonder if there's still some reasonably easy way to verify the inclusions in that case. There are certainly less easy ways, which will be required for a more general eigenvalue solver.

For all isolated eigenvalues, it's already possible to just compute the roots of the characteristic polynomial with Arb, though obviously this is bad with large matrices.

argriffing:
> For my own purposes right now, I need the eigenvalues of the matrix I'm diagonalizing to be unique, and when they aren't I compute a different matrix to diagonalize.

Would it make sense to limit the scope to this case for now? In other words give up if any eigenvalue intervals overlap.

argriffing:
closing as 'inactive'

PR #131, opened 2016-04-14
Add convenient print functions

rickyefarr:
I find myself often implementing print like functions that will print new lines for convenience.  If you agree that this change would be worthwhile I can do this for other types within this merge request.

If you'd like to add these functions, would you like documentation as well?

I named the function arb_println because it is similar to the name java uses which I believe is out_println.

### DIFF
--- a/arb.h
+++ b/arb.h
@@ -277,6 +277,13 @@ arb_print(const arb_t x)
 }
 
 ARB_INLINE void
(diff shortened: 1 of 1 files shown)


rickyefarr:
I will resubmit this pull-request under a different branch than master :)

PR #132, opened 2016-04-14
add convenient print functions

rickyefarr:
I find myself often implementing print like functions that will print new lines for convenience. If you agree that this change would be worthwhile I can do this for other types within this merge request.

If you'd like to add these functions, would you like documentation as well?

I named the function arb_println because it is similar to the name java uses which I believe is out_println.

### DIFF
--- a/arb.h
+++ b/arb.h
@@ -277,6 +277,13 @@ arb_print(const arb_t x)
 }
 
 ARB_INLINE void
+arb_println(const arb_t x)
+{
+    arb_fprint(stdout, x);
+    printf("\n");
+}
+
+ARB_INLINE void
 arb_printd(const arb_t x, slong digits)
 {
     arb_fprintd(stdout, x, digits);


fredrik-johansson:
IMHO typing printf("\n"); isn't that much trouble. In isolation, this could be a convenient addition, but there are already too many print functions: print, printd, printn, fprint, fprintd, fprintn. For consistency, there should arguably be ln versions of all of them, if there is for any one of them.

rickyefarr:
That's why I didn't implement the rest, I wanted your opinion first.  I agree that printing a new line would be needed for all the other functions as well.

argriffing:
> there are already too many print functions

If you think there are too many, would it make sense to remove some? As the project grows I imagine that backwards compatibility concerns will make it more difficult to delete them.

fredrik-johansson:
I think it's already large enough that backwards compatibility is a concern, especially for basic functions like printing.

argriffing:
The fprint* functions were added more recently. They could be combined into one function?

Issue #36, opened 2015-01-16
Add incomplete gamma functions


fredrik-johansson:
Gamma(s,z)
gamma(s,z)
gamma*(s,z) (without the z^s factor)

Series expansions with respect to z.

Maybe also regularized incomplete gamma functions.

argriffing:
At least some of these have been added, right?

fredrik-johansson:
Yes, the upper gamma function (Gamma(s,z)) has been added, including regularization. It's missing series expansion (there is a function in acb_poly, but it should be moved to acb_hypgeom and a regularized flag should be added).

The lower gamma function is missing. It's a trivial special case of 1F1, just not with code yet. To keep the number of functions down, I think I would make gamma*(s,z) the result of gamma(s,z,regularized) with regularized=2.

PR #133, opened 2016-04-16
ENH: add lower gamma

argriffing:
This is mostly copied from the upper gamma implementation. I don't know about a couple of the edge cases that were specially handled by the upper gamma code, so I've left some 'todo' comments. Maybe it's related to branch cuts or something, I'm not sure.

It partially addresses https://github.com/fredrik-johansson/arb/issues/36.

### DIFF
--- a/doc/source/acb_hypgeom.rst
+++ b/doc/source/acb_hypgeom.rst
@@ -589,15 +589,16 @@ Incomplete gamma functions
     The *singular* version evaluates the finite sum directly and therefore
     assumes that *s* is not too large.
 
-.. function:: void acb_hypgeom_gamma_lower_1f1a(acb_t res, const acb_t s, const acb_t z, int regularized, slong prec)
-
-.. function:: void acb_hypgeom_gamma_lower_1f1b(acb_t res, const acb_t s, const acb_t z, int regularized, slong prec)
-
 .. function:: void acb_hypgeom_gamma_lower(acb_t res, const acb_t s, const acb_t z, int regularized, slong prec)
 
-    Computes the lower incomplete gamma function `\gamma(s,z)`
-    with regularization options similar to those for the upper
-    incomplete gamma function.
+    If *regularized* is 0, computes the lower incomplete gamma function
+    `\gamma(s,z)`.
+
+    If *regularized* is 1, computes the regularized lower incomplete
+    gamma function `P(s,z) = \Gamma(s,z) / \Gamma(s)`.
+
+    If *regularized* is 2, computes a further regularized lower incomplete
+    gamma function `\gamma^{*}(s,z) = z^{-s} P(s,z)`.
 
 Exponential and trigonometric integrals
 -------------------------------------------------------------------------------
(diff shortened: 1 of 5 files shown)


fredrik-johansson:
This is not using asymptotic expansions, and doesn't support negative integer s. The easiest solution would be to just wrap acb_hypgeom_m.

It's not documented what regularized = 2 does; note that this is not actually the same thing that gamma_upper computes with regularized = 2.

argriffing:
OK I see what you mean, the implementation should look more like http://dlmf.nist.gov/8.5 where `M` is `acb_hypgeom_m`.

> It's not documented what regularized = 2 does; note that this is not actually the same thing that gamma_upper computes with regularized = 2.

If I'm reading the dlmf correctly, I guess that for lower gamma the `s^-1` and `z^s` terms should both be skipped when regularized=2. For the upper gamma only the `z^s` term is skipped.

fredrik-johansson:
Yes, except that regularized=2 in gamma_upper is not really meant to be used by directly; it's just an internal feature that allows expint to be implemented conveniently. It could be renamed to regularized=-1 or regularized=271 to avoid confusion with regularized=2 for gamma_lower, which is a named function in its own right.

doc/source/acb_hypgeom.rst
```diff
@@ -589,6 +589,17 @@ Incomplete gamma functions
     The *singular* version evaluates the finite sum directly and therefore
     assumes that *s* is not too large.
 
+.. function:: void acb_hypgeom_gamma_lower(acb_t res, const acb_t s, const acb_t z, int regularized, slong prec)
+
+    If *regularized* is 0, computes the lower incomplete gamma function
+    `\gamma(s,z)`.
+
+    If *regularized* is 1, computes the regularized lower incomplete
+    gamma function `P(s,z) = \Gamma(s,z) / \Gamma(s)`.
```
fredrik-johansson: \gamma(s,z)

doc/source/acb_hypgeom.rst
```diff
@@ -589,6 +589,17 @@ Incomplete gamma functions
...
+
+    If *regularized* is 2, computes a further regularized lower incomplete
+    gamma function `\gamma^{*}(s,z) = z^{-s} P(s,z)`.
```
fredrik-johansson: Missing the / s. Could also write that this is equal to M(...).

argriffing:
I've updated this PR to just wrap acb_hypgeom_m, but I'm not sure if the details related to singularities, holomorphic extensions, or branch cuts are correct. Some DLMF identities are tested.

doc/source/acb_hypgeom.rst
```diff
@@ -589,6 +589,17 @@ Incomplete gamma functions
     The *singular* version evaluates the finite sum directly and therefore
     assumes that *s* is not too large.
 
+.. function:: void acb_hypgeom_gamma_lower(acb_t res, const acb_t s, const acb_t z, int regularized, slong prec)
+
+    If *regularized* is 0, computes the lower incomplete gamma function
+    `\gamma(s,z)`.
```
fredrik-johansson: Could give the definition in terms of 1F1 here.

doc/source/acb_hypgeom.rst
argriffing: fixed

doc/source/acb_hypgeom.rst
argriffing: I've added this, copying some markup from the upper gamma docs.

PR #134, opened 2016-04-19
MAINT: move upper incomplete gamma series from acb_poly to acb_hypgeom

argriffing:
part of https://github.com/fredrik-johansson/arb/issues/36

### DIFF
--- a/acb_hypgeom.h
+++ b/acb_hypgeom.h
@@ -126,6 +126,9 @@ void acb_hypgeom_gamma_upper_1f1b(acb_t res, const acb_t s, const acb_t z, int m
 void acb_hypgeom_gamma_upper_singular(acb_t res, slong s, const acb_t z, int modified, slong prec);
 void acb_hypgeom_gamma_upper(acb_t res, const acb_t s, const acb_t z, int modified, slong prec);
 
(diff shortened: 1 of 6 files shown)


argriffing:
fixed docs typo that had been present in the acb_poly docs (upper_gamma_series -> gamma-upper_series)

fredrik-johansson:
The new version should have a regularized flag.

Later, there should also be a version of this function taking a series for the parameter. But that can wait.

argriffing:
> The new version should have a regularized flag.

Added this.

fredrik-johansson:
Thanks.

The regularized flag should go before n in the argument list.

In the regularized==1 branch, you can reuse t, u or v instead of creating c.

It's better to multiply by rgamma than dividing by gamma (consider s being a negative integer).

I'm confused by what happens in the regularized==2 branch. It seems to multiply by exp(-h(x)), which shouldn't give the right thing...

argriffing:
> in the regularized==2 branch

Oops, it should be changed to z^-s instead of e^-z, right? I've made these changes.

fredrik-johansson:
Aliasing seems to be broken now, because g is written to before the last time h is read from. Why does it pass the test code?

argriffing:
> Why does it pass the test code?

The failure mode was nan+/-inf which passes the overlap test

fredrik-johansson:
Okay, I think this should be fine now.

Issue #135, opened 2016-04-21
Build extension module examples when using 'make examples'


rickyefarr:
It would be convenient to automatically build examples from any extension modules, using make examples.  So, if example.c was placed within an examples folder inside the extension module, the program is automatically compiled as part of arb.

rickyefarr:
It could be useful to add this to Flint as well, especially if compiling arb as an extension module.

rickyefarr:
This was fixed with the latest commit.

Issue #113, opened 2016-03-10
arb_sqr


argriffing:
This could just be a small wrapper around `arb_mul(b, a, a, prec)` for now.

rickyefarr:
Added arb_sqr and acb_sqr as well as documentation in commit:

[](https://github.com/rickyefarr/arb/commit/73011a5386120dd3fa20215bff4df4c757f06b3e)

rickyefarr:
Should other types have a squaring method?

fredrik-johansson:
Maybe. There are already squaring methods for matrices. With polynomials, I guess there should also be sqrlow, and for arb_t/acb_t, addsqr, subsqr.

rickyefarr:
I'll add these when I get the chance, but I do have a question.  Are there any particular reason why we wouldn't use the preprocessor to define these functions?  For example, instead of using an inlined function, use

#define arb_sqr(res, x, prec) arb_mul((res), (x), (x), (prec))

Are there any reasons why this would not be preferable?

rickyefarr:
I haven't done any tests in awhile, but the last time I did mpfr was faster when squaring.  Perhaps the algorithm that is used by mpfr should be implemented?

rickyefarr:
Any updates on this?  Is this worth pursuing further?

fredrik-johansson:
About the preprocessor: the problem is that it doesn't create a linkable function.

Squaring uses mpn_mul and mpfr_mul depending on the precision, so it should be close to optimal. However, there's still some overhead than necessary in arb_mul, which I plan to remove some time...

PR #136, opened 2016-04-21
Build Extension Module Examples, add arb_sqr and acb_sqr

rickyefarr:
This fixes issue #135, although not an elegant fix.  

Add function arb_sqr requested in issue #113  as well as acb_sqr and documentation.


fredrik-johansson:
Seems ok. Thanks. I don't really want to encourage use of the extensions mechanism, but it shouldn't interfere with anything else, so that's fine.

PR #140, opened 2016-04-25
ENH: add incomplete gamma lower series

argriffing:
This is mostly copypasted from gamma upper series. For regularized=0 or regularized=1, higher order coefficients in the lower series should be negatives of the corresponding coefficients in the upper series. Regularization option 2 is defined slightly differently for lower vs. upper gamma, by a factor of Gamma(s).

### DIFF
--- a/acb_hypgeom.h
+++ b/acb_hypgeom.h
@@ -131,6 +131,9 @@ void acb_hypgeom_gamma_upper_series(acb_poly_t g, const acb_t s, const acb_poly_
 
 void acb_hypgeom_gamma_lower(acb_t res, const acb_t s, const acb_t z, int modified, slong prec);
 
+void _acb_hypgeom_gamma_lower_series(acb_ptr g, const acb_t s, acb_srcptr h, slong hlen, int regularized, slong n, slong prec);
+void acb_hypgeom_gamma_lower_series(acb_poly_t g, const acb_t s, const acb_poly_t h, int regularized, slong n, slong prec);
+
 void acb_hypgeom_expint(acb_t res, const acb_t s, const acb_t z, slong prec);
 
 void acb_hypgeom_erf_propagated_error(mag_t re, mag_t im, const acb_t z);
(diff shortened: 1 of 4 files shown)


argriffing:
As far as I know, this finishes addressing https://github.com/fredrik-johansson/arb/issues/36 (I hope that phrasing was oblique enough to not trigger github's auto-close feature!). Or maybe it's not finished if you want to extend these functions to take a series for `s`.

fredrik-johansson:
With regularized=2, it looks like there's no special handling of nonpositive integer s. I, that case, the result should simply be gamma*(-n,h(z)) = h(z)^n.

argriffing:
> With regularized=2, it looks like there's no special handling of nonpositive integer s.

In the process of adding tests for this, I noticed that the plain (non-series) lower gamma evaluation is not robust near `-n` due to `acb_hypgeom_m(..., regularized=1, ...)` giving `nan +/- inf` at intervals like
```
a = -3 +/- eps
b = -2 +/- eps
z = -2.0
```
for small but nonzero `eps`. The implementation currently special-cases `eps=0` and gives an informative result in that limit. According to DLMF this regularized M function should be entire in `z`, `a`, and `b`. That means it should be possible to improve the implementation so that it does not give `nan +/- inf` at these tiny intervals, right?

fredrik-johansson:
Correct. This is a problem affecting many other functions as well.

By the way, you can update the copyright/license declarations to the new template now.

argriffing:
> In that case, the result should simply be gamma*(-n,h(z)) = h(z)^n.

OK I've added this with a test.

fredrik-johansson:
The case should be dealt with in the underscore method. Also, you could call _acb_poly_pow_acb_series, since that method handles integers, with less fiddling involved. (But a temp allocation will be needed to handle aliasing.)

Please update the license declarations, too.

argriffing:
I've made those changes.

fredrik-johansson:
looks good now, thanks!

PR #141, opened 2016-04-28
ENH: acb poly scalar mul and div convenience functions

argriffing:


fredrik-johansson:
These are overdue. How about arb_poly counterparts?

fredrik-johansson:
I noticed after merging that the methods didn't normalise the output, and fixed this in the master. Normalising is necessary after multiplying by 0 or dividing by inf.

PR #143, opened 2016-05-02
ENH: arb_is_indeterminate

argriffing:
This PR would add a helper function `arb_is_indeterminate`. I'm not sure if it's named appropriately (should be `arb_is_nan`?) or if the idea is even sound.


fredrik-johansson:
Let me think about this. I'm not entirely convinced that distinguishing between different kinds of non-finite intervals is useful enough to be encouraged by more explicit interfaces (at least yet)... the treatment is very unsystematic elsewhere.

argriffing:
closing as 'inactive'

Issue #147, opened 2016-06-03
arb_mat_exp entrywise relative accuracy


argriffing:
Currently the exponential of a tiny-in-norm matrix gives `identity +/- eps`, which has no entrywise relative accuracy for off-diagonal entries. Maybe this makes sense because in general arb does not try to control entrywise relative error, but I had to patch this to make a piece of my code work.

```diff
slong
 _arb_mat_exp_choose_N(const mag_t norm, slong prec)
 {
-    if (mag_is_special(norm) || mag_cmp_2exp_si(norm, 30) > 0 ||
-        mag_cmp_2exp_si(norm, -prec) < 0)
+    if (mag_is_special(norm) || mag_cmp_2exp_si(norm, 30) > 0)
     {
         return 1;
     }
+    else if (mag_cmp_2exp_si(norm, -prec) < 0)
+    {
+        /*
+         * 2 Taylor terms give the off-diagonal entries a chance
+         * to have finite relative error.
+         */
+        return 2;
+    }
```

This isn't a PR because I understand that in general it might not be feasible or desirable to try to control entrywise relative accuracy of matrix functions. Maybe this could be addressed by a function like `arb_mat_expm1`? For that hypothetical function the attempt to control the error norm rather than error entries would have been enough for me.

fredrik-johansson:
The patch looks like a good change to me, even if it's not a completely general solution.

PR #148, opened 2016-06-03
MAINT: improve entrywise relative accuracy of exponentials of matrices with small norms

argriffing:
closes https://github.com/fredrik-johansson/arb/issues/147


fredrik-johansson:
thanks

Issue #154, opened 2016-07-19
Implement Fused Multiply Add and Sub


rickyefarr:
Currently, there is a similar function arb_addmul(acb_t z, const acb_t x, const acb_t y, slong prec).  This computes z = z + x*y.  It would be nice to implement a function that is similar to MPFR's where the output is w=x*y + z.

Shouldn't be too hard to adjust the current implementation of arb_addmul to do this.

rickyefarr:
I have successfully created arf_fam (reverse of mpfr version, w=op1 + op2*op3).  I also created a test function.  As suspected, it's very easy to change the existing addmul code to create these functions.

If you think this would be useful to include in the library, let me know.  I know that it is useful for me at least.

PR #156, opened 2016-08-06
Add arf_fam function

rickyefarr:
This function is for fused addition and multiplication for the arf type.  This is part of a fix for issue #154


rickyefarr:
Don't pull this in.  It appears as though I'll need to add the same variable that arf_addmul has so that this works as expected.

rickyefarr:
For the time being, I will close this pull request and re-request when this issue is fixed fully.

fredrik-johansson:
It would make sense to name it arf_fma and copy the MPFR operand order.

Also, it's probably not necessary to have separate implementations of fma and addmul since the code essentially is identical. addmul could just be changed to an inline function that invokes fma.

rickyefarr:
Alright, I'll fix the code and update this.  The only part that may give me issues is the adaptation of the test module.

PR #157, opened 2016-08-06
Remove unnecessary variable, tptr2, in arf_addmul, arf_addmul_mpz, arf_submul, and arf_submul_mpz

rickyefarr:
It appears as though the variable tptr2 was unnecessary.  Probably included accidentally from testing the algorithm and not removed.

### DIFF
--- a/arf/addmul.c
+++ b/arf/addmul.c
@@ -16,7 +16,7 @@ arf_addmul(arf_ptr z, arf_srcptr x, arf_srcptr y, slong prec, arf_rnd_t rnd)
 {
     mp_size_t xn, yn, zn, tn, alloc;
     mp_srcptr xptr, yptr, zptr;
-    mp_ptr tptr, tptr2;
+    mp_ptr tptr;
     fmpz_t texp;
     slong shift;
     int tsgnbit, inexact;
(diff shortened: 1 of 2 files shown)


rickyefarr:
Well, that was shocking.  It passes the tests for arf_addmul, but the change causes strange behavior within the test module for arb_const_pi.

I didn't know that it failed these tests until travis-ci caught it, since I only checked that it passed the arf tests.   Surprising that it passed so many tests before it failed with this particular one.

rickyefarr:
I will close this pull request, but I'd really like to discuss this.  Judging by the code and the tests for arf type variables, this should work.  

The tests involving the arb type shouldn't catch this should it?  Are the arf tests not robust enough?

rickyefarr:
Okay, looking through the source code...there are only 9 places where ARF_MUL_TMP_ALLOC is called.  

I *think* in my own head, I understand why the tptr2 variable is required.  What I'm curious about is, wouldn't it be best to hide the need for this variable within the pre-processor code?

So, for example, I propose that ARF_MUL_TMP_ALLOC(tmp, alloc) declare a pointer by the name tmp__2, or something similar using the pre-processor's magic ## symbol.  Do all the current commands on the variable tmp__2, and then at the end of the function put tmp=tmp__2. (Maybe without a semi-colon so that the function should be called with a semi-colon at the end.

Does this make sense?  IMHO, I think that hiding the fact that the tptr2 variable is necessary makes the code more readable...because just by reading addmul, it's not intuitive that that is the case.

PR #160, opened 2016-08-16
Move some existing documentation for acb types

rickyefarr:
It's a matter of opinion, but I think that the documentation for acb_is_zero, acb_is_one, acb_is_finite, and acb_is_exact should be in section "Precision and Comparisons" and not in the section "Basic Manipulations".  If you disagree, I'll close this pull request.


fredrik-johansson:
Seems better.

PR #161, opened 2016-08-16
Add Complex Versions of the falling factorials

rickyefarr:
In a later pull request, I'll add the arb version of the falling factorials.  Also, I'll add a falling version of acb_poly_rising_ui.

I didn't know if the functions should be inlined or not.  I'm assuming they should be since they are very short.  Then again, the clear functions were short too but eventually un-inlined.

I added a new file, falling.c, that includes only the function acb_falling.  It seems like acb_falling and acb_rising should be together in the same file...maybe factorial.c?

### DIFF
--- a/acb.h
+++ b/acb.h
@@ -768,11 +768,104 @@ void acb_rising_ui_rec(acb_t y, const acb_t x, ulong n, slong prec);
 void acb_rising_ui(acb_t z, const acb_t x, ulong n, slong prec);
 void acb_rising(acb_t z, const acb_t x, const acb_t n, slong prec);
 
+ACB_INLINE void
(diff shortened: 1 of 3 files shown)


fredrik-johansson:
It's not necessary to wrap the different implementation variants (bs, etc.). acb_falling and acb_falling_ui should be enough.

rickyefarr:
I'll remove those functions and update the documentation soon.

fredrik-johansson:
This seems stale enough to close.

Again, I would be fine with adding arb_falling_ui/acb_falling_ui if they are implemented as simple wrappers of the rising factorial functions, but the other versions seem completely unnecessary.

PR #165, opened 2016-09-10
Dirichlet

pascalmolin:
base PR for Dirichlet module, thanks for extending l_hurwitz.

For L-functions, I am working on l_incgam (simple) and then I will add the Fourier method.


pascalmolin:
I have a segfault with example/lcentral --prec 128 3 5
(whenever prec value is >= 128)
I did not understand where it comes from.

fredrik-johansson:
In _acb_dft_cyc, there is this code:

        for (i = 0; i < m; i++)
            for (j = 0; j < M; j++)
                w[j + M * i] = t[i + m * j];
        _acb_vec_clear(t, m * M);

The assignments should presumably be acb_swaps.

I have not yet checked if there are other such instances.

pascalmolin:
Well, I no longer need the t array (temp variables) so I just copy to w.
The same code occurs in _acb_dft_prod.
I also have similar non standard code in the _vec_extract function with
acb_struct copies.
It could be acb_set, if it is the same.

By the way, all this comes from reordering needs, I can do better in this
file, but I deliberately left
the code operating close to my gp scripts for debugging purposes.
In particular, the _dft_prod and _dft_cyc function are very similar, they
should be merged eventually
(the butterfly step can be parametrized by one common dft struct).

2016-09-10 17:41 GMT+02:00 Fredrik Johansson <notifications@github.com>:

fredrik-johansson:
When you do the shallow copy w[j] = t[i], it has two bad effects:

1) overwrites any pointers already allocated in w[j], so there is a memory leak
2) since w now contains the same pointer, clearing the vector t will free the data in t[i], which means that the data in w[j]. This leads to a segfault later when you try to free w (and possibly earlier when you try to read from it), since this data has already been freed and cannot be freed again.

pascalmolin:
Oh, I see... If I understand well there is no problem in the _vec_extract
function, but I have to replace these w[xx] = t[yy] by acb_set(w + xx, t +
yy).
Thanks!

2016-09-10 18:10 GMT+02:00 Fredrik Johansson <notifications@github.com>:

fredrik-johansson:
BTW, it might be better to develop the incomplete gamma and Fourier code in a new branch so that I can review and merge the basic features more quickly. I don't want to merge incomplete code (missing error bounds, disabled tests, etc.).

fredrik-johansson:
- Many of the minor dlog methods are missing documentation. It's not a big problem since they are not meant to be called directly by users, but would be helpful if someone else who wants to understand the algorithms.

- dlog_precomp_n_init calls n_is_probabprime and not n_is_prime -- is the code correct if n is composite?

fredrik-johansson:
.. function:: int acb_dirichlet_conrey_eq(const acb_dirichlet_group_t G, const acb_dirichlet_conrey_t x, const acb_dirichlet_conrey_t y)

        Return 1 if *x* equals *y*. This function checks both *number* and *index*, writing ``(x->n == y->n)`` gives a faster check.

Why not just the faster check in the code?

fredrik-johansson:
There are also still some undocumented methods in the acb_dirichlet module.

fredrik-johansson:
Some of the dlog code fails on 32-bit:

    vec....FAIL: log(137,6) mod 893634271: eratos->863388924 != sieve->142958708

pascalmolin:
2016-09-14 14:07 GMT+02:00 Fredrik Johansson <notifications@github.com>:

>
>    -
>
>    Many of the minor dlog methods are missing documentation. It's not a
>    big problem since they are not meant to be called directly by users, but
>    would be helpful if someone else who wants to understand the algorithms.
>
> I can add documentation, OK.


>
>    -
>
>    dlog_precomp_n_init calls n_is_probabprime and not n_is_prime -- is
>    the code correct if n is composite?
>
> Yes (it just misses the crt decomposition).

pascalmolin:
I think I should use the faster code indeed, and rename this one
`conrey_eq_deep` or something like that (I use it in the tests to check the
integrity of conrey elements).

2016-09-14 14:21 GMT+02:00 Fredrik Johansson <notifications@github.com>:

pascalmolin:
> There are also still some undocumented methods in the acb_dirichlet module.
>
>
As for the dlog module, I thought it would be preferable to document only
methods considered as public, but I can document everything.
I am also writing some notes with more details and proofs of all the
formulas used.

pascalmolin:
I'm looking into it. BTW, is there a simple way to do a 32 bits check? (I
have only access to 64bits machines)

Pascal

2016-09-14 14:53 GMT+02:00 Fredrik Johansson <notifications@github.com>:

fredrik-johansson:
Note: I've pushed a number of other minor fixes directly on my branch, and will continue to do so while reviewing.

> I think I should use the faster code indeed, and rename this one `conrey_eq_deep` or something like that (I use it in the tests to check the integrity of conrey elements).

That's fine. Add a note that it's a function intended for testing, and there should be no confusion.

> As for the dlog module, I thought it would be preferable to document only methods considered as public, but I can document everything. I am also writing some notes with more details and proofs of all the formulas used.

You can put all the internal methods in their own section and warn readers not to use them. It's always useful to give a precise statement of what a function does, particularly whether there are any implicit assumptions on the inputs.

> I'm looking into it. BTW, is there a simple way to do a 32 bits check? (I have only access to 64bits machines) 

A relatively easy way is to set up a 32-bit Linux machine in VirtualBox (this is what I've done).

fredrik-johansson:
- There are a couple of "todo" comments currently in the code left to fix.

Other than the previous comments, it looks pretty good now.

fredrik-johansson:
I found a minor problem. The function n_mulmod2 is a recent addition to flint; for the time being, it's best to keep arb compatible with the flint-2.5.2 release.

It is only used in two places: ui_pairing_conrey and conrey_pow. In both cases, it actually appears in a loop, so precomputing the inverse outside the loop and calling n_mulmod2_preinv will save a few nanocycles anyway.

PR #176, opened 2017-02-28
Replace abort with flint_abort.

thofma:
This will allow us to not loose the julia session in case we hit an error in arb. We did the same for flint, see
https://github.com/wbhart/flint2/pull/243

~~I did not run `make check`, since it is failing for me for other reasons (see the flint-devel mailing list).~~


fredrik-johansson:
Travis is happy.

PR #179, opened 2017-03-02
Change abort to flint_abort only if flint > 2.5.2

thofma:
Since there is no global header file (or I could not find it), these checks are scattered all over the place.


fredrik-johansson:
OK, let's see how this goes. Thanks.

PR #187, opened 2017-05-21
install static library in LIBDIR instead of lib

kiwifb:
Currently `Makefile.in` installs static libraries in `$(DESTDIR)/$(PREFIX)/lib` instead of `$(DESTDIR)/$(PREFIX)/$(LIBDIR)` as the dynamic one. This can have funny consequence since `$(DESTDIR)/$(PREFIX)/$(LIBDIR)` is created by the install target but `$(DESTDIR)/$(PREFIX)/lib` isn't. `/usr/lib` usually exist on unix but when `DESTDIR` is set all your bets are off. As seen in https://github.com/cschwan/sage-on-gentoo/issues/471 the side effects are interesting.


fredrik-johansson:
Thanks!

PR #191, opened 2017-10-07
cmake: find libraries with no _lib suffix

xoviat:
Some libraries (particularly mpir) don't have a lib_ prefix.

If I understand correctly, this should fix the problem.


xoviat:
Actually this did not work for me.

PR #195, opened 2017-10-20
Dft

pascalmolin:
The memory leaks should be fixed, more documentation.

PR open for discussion.


fredrik-johansson:
Thanks! I will review this closely soon.

Comments:
* Could convolution be computed via _acb_poly_mul/_acb_poly_mullow?
* Cooley-Tuckey should be Cooley-Tukey
* Wrappers for DFT/IDFT of real data as well as DCT would be nice (but not needed for this PR)
* The test for inverse DFT seems to be commented out
* The test code is not trying random lengths. Some basic random tests of e.g. DFT <=> IDFT would be good. Also make sure length 0 and 1 are tested.

pascalmolin:
I do not think convolution can be computed with poly_mullow, that would be inefficient (doubling lengths and taking only half of the result).
I added the requested tests.

One issue + interface question: right now the implementation does not correspond to the documentation, it computes with exp(2*I*pi/n) instead of exp(-2*I*pi/n) [it does not matter but this is the usual definition]. I see two options:
- hardchange the code (basically compute _acb_vec_unit_roots_conj)
- have a flag to choose the sign at runtime in each internal function, this makes pointer access more complicated but simplifies the dft_inverse functions (no reordering).
Do you have a preference/idea?

Other comment: I do not think dft_cyc is optimal in any sense. The dv shift argument is to avoid reordering of the input or the output, this is not cache friendly but I know nothing about this and I think this is OK right now for arb-type computation.

fredrik-johansson:
I think it would be best to change it to the "usual definition".  I think I slightly prefer just hardchanging the code over adding flags. BTW, _acb_vec_unit_roots computes more roots than needed; there's no need to store the ones that can be recovered by negation and conjugation. This shouldn't affect speed much, but it wastes some memory for very long transforms.

It would still be interesting to profile convolution via poly_mullow. It can be faster with low precision; see slide 13 of http://fredrikj.net/math/arith2017.pdf -- at 100 bits precision polynomial multiplications via FLINT is more than 2x faster than via floating-point FFT, and that's while actually doing 12 multiplications internally (4 real multiplications, and each real multiplication uses 1 main mul + 2 for error bounds). For DFT computations we don't need the best relative error for all coefficients as in polynomial multiplication (I think); only the global absolute error really matters. So I think we could do something like this: scale the inputs so that the largest coefficient is ~= 2^(prec+log(n)+O(1)), truncate to fmpz vectors and multiply via _fmpz_poly_mullow (using Gauss/Karatsuba formula to do one complex multiplication via three real ones). Compute error bounds in O(n) operations just using norms.

The code looks good. I just noticed that there's a .tmp file left over

fredrik-johansson:
The tests fail on Travis, by the way. One of the configurations always fails nowadays due to some build issue, but there is a failing dft test on the other ones.

fredrik-johansson:
For the failure in t-convol, acb_dft_convol_rad2_precomp and acb_dft_convol_rad2 need to check for len == 0

t-dft takes a bit too long (20 seconds). 2-3 seconds should be enough. Reduce the random length slightly and avoid using dft_naive for long transforms.

I think the arguments order and len to vec_unit_roots should be reversed.

pascalmolin:
Thanks for the comments Fredrik.

- I'm glad you accept the extra len argument for _acb_vec_unit_roots :-) I can swap the arguments if you think this is the right thing to do, I must say wanted to mimic the usage I guessed from other _acb_vec functions: length first.
- the convol_mullow function is quite stupid right now because length is doubled and maybe doubled again in mullow, I guess one could branch directly to some internal flint convolution as you proposed. I must say I did not think about it, I think it is worth asking Bill Hart.
- acb_dirichlet_dft gets its conjugate, this is not very convenient for usual sums. Anyway I think a next step is to implement general character sums, the vec variants of which will give an entry point to non-conjugate dft.

fredrik-johansson:
The output length is generally supposed to come last (before prec). The general pattern is:

output vector, [input vector, input length, ...], [parameter], output length, prec

The exception in _arb_vec_scalar_mul for example is that you give the input length, and since this is the same as the output length, the output length is not required separately. This might not be 100% consistent, but I do think length last is correct here.

fredrik-johansson:
I think the profiler output should probably not be part of the repository...?

pascalmolin:
Profile output should be discarded before merging, I just wanted to discuss these timings.
It appears that precomputing gains almost nothing except for bluestein. I'm a bit surprised, does this mean that memory access is already horrible for small size?

fredrik-johansson:
> Profile output should be discarded before merging, I just wanted to discuss these timings.

OK. It just adds a bit of noise to the commit history. A good place to put profile data and the like is on the wiki.

> It appears that precomputing gains almost nothing except for bluestein. I'm a bit surprised, does this mean that memory access is already horrible for small size?

I doubt it. Precomputation is just O(n) so it shouldn't have a big impact.

fredrik-johansson:
Attaching files to the PR comments also works, of course.

pascalmolin:
Is it ok if I rebase without the logs or is it safer to just revert?

I agree with the O(n) fact, but for several timings I'm also doing O(n) iterations so I really expected a noticeable impact. But I see that after one crt step the sizes are really small so this is normal.

fredrik-johansson:
> Is it ok if I rebase without the logs or is it safer to just revert?

Your preference.

fredrik-johansson:
Is this ready to be merged?

pascalmolin:
Yes, you can.

PR #203, opened 2018-01-18
Remove mysterious 2 factor.

jpflori:


fredrik-johansson:
Demystification accepted.

jpflori:
Forgot to remove the q <= ULONG_MAX/2 check.

PR #204, opened 2018-01-18
Other simplifications for gauss periods.

jpflori:


fredrik-johansson:
Finally merging this. Thanks!

Issue #198, opened 2017-12-03
Nonzero real part returned from arccosh(z), Im(z)=0, abs(Re(z))<1


rwst:
I think there should be a shortcut in `acb_acosh` that sets any real part of the result to zero if `Im(z)=0, abs(Re(z))<1`, see
https://mathoverflow.net/questions/287602/does-re-operatornamearccoshz-vanish-with-imz-0-1-rez1

At the moment this has nonzero real part:
```
sage: CBF(0).arccosh()
[1.570796326794897 +/- 6.65e-16]*I
sage: CBF(-1/2).arccosh()
[+/- 4.90e-16] + [2.09439510239320 +/- 5.81e-15]*I
sage: CBF(1/2).arccosh()
[+/- 4.90e-16] + [1.047197551196598 +/- 7.94e-16]*I
```

rwst:
I guess a similar argument applies to `acb_acos` for `Im(z)=0, Re(z)>1:
```
sage: CBF(2).arccos()
[+/- 4.45e-16] + [1.316957896924817 +/- 7.09e-16]*I
```

PR #210, opened 2018-03-04
return exact zero where possible: acos, acosh

rwst:
See #198

### DIFF
--- a/acb/acos.c
+++ b/acb/acos.c
@@ -29,9 +29,8 @@ acb_acos(acb_t res, const acb_t z, slong prec)
             arb_one(one);
             if (arb_gt(acb_realref(z), one))
             {
-                acb_zero(t);
                 acb_asin(res, z, prec);
-                acb_sub(res, t, res, prec);
+                acb_neg(res, res);
                 arb_zero(acb_realref(res));
                 arb_clear(one);
                 arb_clear(t);
--- a/acb/acos.c
+++ b/acb/acos.c
@@ -32,6 +32,7 @@ acb_acos(acb_t res, const acb_t z, slong prec)
                 arb_clear(one);
                 return;
             }
+            arb_clear(one);
         }
         acb_t t;
         acb_init(t);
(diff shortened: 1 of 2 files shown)


fredrik-johansson:
The code does not support aliasing properly, as res is written to before the last read of z.

rwst:
Of course, I could simply clear and return after zeroing in acos(). Funny because I just did a clean up of such indentations in pynac...

fredrik-johansson:
It's puzzling why t-acos and t-acosh don't fail since they are supposed to check aliasing.

fredrik-johansson:
The variable one is not always being cleared (this technically doesn't matter since z=1 doesn't result in memory allocation, but it's not elegant).

fredrik-johansson:
Aren't the functions mixed up? acosh should be real for x >= 1, acos should be real for -1 <= x <= 1. And acos is not actually equal to asin... it's a mystery that the test code isn't failing...

rwst:
The results are not real, their real part should be exactly 0 (but isn't), as is proven for acosh in the link given in #198. I don't get your statement with asin, sorry.

fredrik-johansson:
The logic you added to acb_acos does this:

if im(z) == 0 and re(z) >= 1, set res to re(asin(z))

The logic for acb_acosh checks if abs(re(z)) <= 1.

So the comparisons for acb_acos and acb_acosh seem to be switched, and additionally acb_acos is not giving the right value.

rwst:
It is not about where acosh(z) is real, it's about where it is purely imaginary. This is the case with -1<z<1, but acb_acosh() returns a ball around 0.

I see for acos, that asin should have been negated.

rwst:
I meant to say "...but acb_acosh() returns a ball around 0 as real part"

fredrik-johansson:
OK thanks, I see. Though why use acb_sub when acb_neg is sufficient? :-)

rwst:
I replaced with acb_neg . While testing passes on my machine Travis fails consistently in one place. Was Travis testing tested recently?

fredrik-johansson:
It started timing out just now. Perhaps one of the Travis machines is under too heavy load or misconfigured. If it persists, we can lower the number of test iterations for that configuration.

fredrik-johansson:
Thanks!

PR #215, opened 2018-03-11
fix rpath

videlec:


fredrik-johansson:
I will assume that this works until someone complains.

PR #216, opened 2018-03-18
Arb mat solve precond

arbguest:


fredrik-johansson:
Thanks!

I have a hunch that arbguest = Alex Griffing. We can keep the aliases in the copyright lines if you prefer, but is that the right name to put in credits.rst in the documentation and in the changelog?

Issue #164, opened 2016-09-09
Dirichlet L-functions (meta issue and feature wishlist)


fredrik-johansson:
- [ ] Review all code in the dirichlet branch thoroughly and merge into master (PR #163).

- [ ] Add more example programs.

- [ ] Series expansions.
- [ ] Evaluation using reflection formula.
- [ ] Generalized Bernoulli numbers / evaluation at -n? Functions to compute the exact rational coefficients?
- [ ] Euler product evaluation for re(s) >> 0.

- [ ] Move Hurwitz zeta code to the Dirichlet module and tidy the code a bit.
- [ ] Look for ways to optimize Hurwitz zeta for rational parameter.

- [ ] Possible use of double and double-double arithmetic to speed up truncated L-series.
- [ ] Improve parameter selection in Euler-Maclaurin summation and avoid
      unnecessary slowdown for large height.
- [ ] Allow the user to limit the memory allowed for sieving etc.

- [ ] Completed L-functions / Riemann xi-function, Z-function, Siegel theta function, Gram points.
- [ ] Computing zeros.
- [ ] Polynomial approximation with error bounds.
- [ ] Riemann-Siegel formula.
- [ ] Grid evaluation.

PR #218, opened 2018-03-29
add riemann xi function

p15-git-acc:
Add a bad implementation of https://en.wikipedia.org/wiki/Riemann_Xi_function mentioned in https://github.com/fredrik-johansson/arb/issues/164. A better implementation would be defined at s=1 but I don't know how to do it.


fredrik-johansson:
Thanks! Could you make the following changes:

* Rename it to `acb_dirichlet_riemann_xi` (don't forget to rename the file and the test code) to keep the prefix of the acb_dirichlet module consistent. Or maybe just `acb_dirichlet_xi`? (I think there is no other xi function related to Dirichlet L-functions that it can be confused with -- the completed general L-functions are usually called Lambda.)
* Add the documentation to `doc/source/acb_dirichlet.rst`.
* Instead of evaluating (s/2) gamma(s/2), evaluate gamma(1+s/2). Then there is no spurious singularity at s = 0.
* Handling s=1 is easy: do `if (acb_is_one(s)) { /* set res = 1/2 */ }`. This will still not work for non-point interval containing 1, but that can be fixed later.
* Use the functional equation for more accurate evaluation in the left half plane (in particular, this avoids the poles of the gamma function). You can use the same pattern as in acb_dirichlet_zeta: evaluate at 1-s instead of s if `arf_sgn(arb_midref(acb_realref(s))) < 0`.
* I prefer to place all the `acb_init` calls together at the top.

fredrik-johansson:
Great, thanks!

fredrik-johansson:
I put in a solution for intervals around s=1, by simply using the functional equation in that case.

p15-git-acc:
It should be changed to not use the functional equation if `s` contains zero. As it is, `xi(s)` is indeterminate for `s` that contains zero with midpoint in the left half plane, even if the interval is small.

`arf_sgn(arb_midref(acb_realref(s))) < 0` should be updated to `((arf_sgn(arb_midref(acb_realref(s))) < 0) && !acb_contains_zero(s))`

fredrik-johansson:
That isn't the case, unless I'm missing something. I tried this:

        arb_set_str(acb_realref(s), "+/- 1e-6", 53);
        arb_set_str(acb_imagref(s), "+/- 1e-6", 53);
        acb_dirichlet_xi(z, s, 53);
        acb_printn(z, 20, 0); printf("\n");
        acb_set_si(s, e);
        acb_dirichlet_xi(z, s, 53);
        acb_printn(z, 20, 0); printf("\n");

and get this:

    [0.5000 +/- 7.62e-5] + [+/- 4.32e-5]*I
    0.50000000000000000000

p15-git-acc:
Your example's midpoint was 0. The problem only appeared when the midpoint was in the left half plane.

fredrik-johansson:
Yes, I understood after seeing your patch. Thanks again!

PR #220, opened 2018-03-30
xi and zeta near zero

p15-git-acc:
Applying the functional equation to xi and zeta when s includes zero and has its midpoint in the left half plane causes infinite error bounds.


fredrik-johansson:
Thanks!

PR #237, opened 2018-09-03
Add LGTM code quality badges

xcorail:
Hi there!

I thought you might be interested in adding these [code quality badges](https://lgtm.com/projects/g/fredrik-johansson/arb/ci/#badges) to your project. They will indicate a very high code quality to potential users and contributors.
You can also check the alerts discovered by [LGTM](https://lgtm.com/projects/g/fredrik-johansson/arb).

N.B.: I am on the team behind LGTM.com, I'd appreciate your feedback on this initiative, whether you're interested or not, if you find time to drop me a line. Thanks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,8 @@ Bug reports, feature requests and other comments are welcome
 in private communication, on the GitHub issue tracker, or on the FLINT mailing list <flint-devel@googlegroups.com>.
 
 [![Build Status](https://travis-ci.org/fredrik-johansson/arb.svg?branch=master)](https://travis-ci.org/fredrik-johansson/arb)
+[![Code Quality: Cpp](https://img.shields.io/lgtm/grade/cpp/g/fredrik-johansson/arb.svg?logo=lgtm&logoWidth=18)](https://lgtm.com/projects/g/fredrik-johansson/arb/context:cpp)
+[![Total Alerts](https://img.shields.io/lgtm/alerts/g/fredrik-johansson/arb.svg?logo=lgtm&logoWidth=18)](https://lgtm.com/projects/g/fredrik-johansson/arb/alerts)
 
 ## Code example
 


fredrik-johansson:
Thanks for doing this. Static analysis is always nice, though as usual, the "alerts" are mostly false positives:

> This pointer might have type struct <unnamed> (size 96), but the pointer arithmetic here is done with type arb_srcptr (size 48).

Yes, this code deliberately casts from one pointer type to another, which could be unsafe in general but is known to be safe here. Why is there an alert when the cast is completely explicit? How else would you do the cast? Cast to void * first? I think that would just make the code less clean.

> Empty block without comment

Strange thing to complain about. I suppose one could put in /* Nothing to do. */ if some people feel uneasy about empty code blocks... :)

> FIXME comment

Perhaps there are different cultures about this, but I don't see how this warrants an alert. I use FIXME comments to indicate that something could be done more cleanly or efficiently, not as an excuse to put in broken code.

> Large object passed by value

This alert seems reasonable, though it doesn't really matter here since the code in question isn't performance-sensitive.

> Comparison result is always the same

In all cases I could see, these comparisons were left in place to make the code valid even if an earlier condition is removed or changed. Not really an issue.

> Self comparison

This is the result of a macro expansion with repeated parameters. To the human eye, this is obviously fine.

> Sign check of bitwise operation

Known to be safe here, unless I misunderstood the warning.

> Declaration hides parameter

Valid (minor) complaint about code quality.

xcorail:
Hello @fredrik-johansson 

Thanks for the detailed feedback! Much appreciated that you took this time for helping us. 

> Thanks for doing this. Static analysis is always nice, though as usual, the "alerts" are mostly false positives:
> 
> This pointer might have type struct (size 96), but the pointer arithmetic here is done with type arb_srcptr (size 48).
> 
> Yes, this code deliberately casts from one pointer type to another, which could be unsafe in general but is known to be safe here. Why is there an alert when the cast is completely explicit? How else would you do the cast? Cast to void * first? I think that would just make the code less clean.
> 

The problem reported here is not with the cast but with the pointer arithmetic, which is known to be unsafe. This is referenced as [CWE-468](https://cwe.mitre.org/data/definitions/468.html).  

> Empty block without comment
> 
> Strange thing to complain about. I suppose one could put in /* Nothing to do. */ if some people feel uneasy about empty code blocks... :)
>

As stated in the help of the alert, if the code is useless, it's always better to remove it. However, note that the severity of this alert is "recommendation".   
 
> FIXME comment
> 
> Perhaps there are different cultures about this, but I don't see how this warrants an alert. I use FIXME comments to indicate that something could be done more cleanly or efficiently, not as an excuse to put in broken code.
> 

Oh no, we agree on this. The problem is not that you put a FIXME comment, nor that you put in broken code. At the contrary, if you put this comment it's to indicate that something better could be done, and the alert is kind of **just showing** this indication in the LGTM front-end. And with severity =  "recommendation", which is consistent.

> Large object passed by value
> 
> This alert seems reasonable, though it doesn't really matter here since the code in question isn't performance-sensitive.
> 

Ok. If you are sure that this code is not (and will never be) performance-sensitive, then you can suppress this alert with a `//LGTM` comment on this line. 
However, the philosophy of the LGTM alerts is not to show you only the problems that you have **now** in your code, but all problems that you could have in the future, for example, if a future contributor makes a change that uses this code in a performance-sensitive context. (I take this as an example, this is perhaps a stupid one in your case, just want to explain the general philosophy of these alerts: account for the future, not only for today).    

> Comparison result is always the same
> 
> In all cases I could see, these comparisons were left in place to make the code valid even if an earlier condition is removed or changed. Not really an issue.
> 

Ok. If this is intended, then you can suppress the alerts. Sometimes, this useless comparisons might indicate a real bug.  

> Self comparison
> 
> This is the result of a macro expansion with repeated parameters. To the human eye, this is obviously fine.
> 

hmmmm ... weird ...  Thanks! I will pass this example to my language team. 

> Sign check of bitwise operation
> 
> Known to be safe here, unless I misunderstood the warning.
> 

The details of the warning are [here](https://lgtm.com/rules/2155680694/). The problem occurs only in edge cases.    

> Declaration hides parameter
> 
> Valid (minor) complaint about code quality.

We agree. Severity = recommendation

I think that the alerts above are mostly showing correct results (except the one with the self comparison), ... however results you don't care about. 
* First, we agree on the definition of false positive here: they are simply alerts that you won't fix. 
* Second, this is not very surprising, as you don't have any "Error", only "Warnings" and "Recommendations", which are less urgent to fix, and always more subject to debates :-)
* How we rank our alerts: from the analysis of 16M commits on github.com, we derive what alerts developers tend to fix quickly, or not quickly, or not at all ... The alerts classification is derived from this data mining. 
* You can suppress alerts, either [globally](https://lgtm.com/help/lgtm/showing-hiding-query-results), or [for a specific line](https://lgtm.com/help/lgtm/alert-suppression)   

Thanks again!
Cheers

xcorail:
Hi @xcorail 

The false positive _"self comparison"_ is being addressed in https://github.com/Semmle/ql/pull/405 
 
Thanks again for reporting it!

fredrik-johansson:
Cool, thanks for the clarifications.

> The problem reported here is not with the cast but with the pointer arithmetic, which is known to be unsafe. This is referenced as CWE-468.

Perhaps I'm still missing something... the CWE-468 that you linked to describes doing arithmetic with the wrong pointer size due to casting to the wrong pointer type. This is *not* what is going on here. The cast and the resulting pointer arithmetic is deliberate. The original pointer refers to an array of n complex numbers, where each complex number is actually a length-2 array of real numbers. The cast is used to reinterpret the array of complex numbers as a length-2n array of real numbers, in order to iterate over the individual real/imaginary components, and the pointer arithmetic is then simply done on offsets for real numbers.

> As stated in the help of the alert, if the code is useless, it's always better to remove it. However, note that the severity of this alert is "recommendation".

Normally, yes, it is better to omit useless if-cases entirely, but it happens to make sense in this case. It would indeed be clearer with a comment!

PR #256, opened 2019-02-06
functions related to zeta zeros

p15-git-acc:


p15-git-acc:
These are low accuracy implementations of S(t) and N(t), with the plan of improving them later with Turing's method.

fredrik-johansson:
Thanks. The interface looks good, but I'm not so sure about merging placeholder implementations yet.

Note: the bound on S could be implemented entirely using mag functions (mag_log, etc).

PR #258, opened 2019-02-08
add backlund_s_bound

p15-git-acc:
Add an upper bound for `|S(t)|` when `t >= 0`. It's related to the irregularity of spacing of nontrivial zeros of the Riemann zeta function.


fredrik-johansson:
Expertly written patch! Thanks!

PR #259, opened 2019-02-10
add turing_method_bound

p15-git-acc:
Add an integer bound related to Turing's method, like sure_number_block in zetazeros.py in mpmath. gram_point is generalized following Brent 1979 to allow n=-1. Tests and docs are added/updated.


fredrik-johansson:
Thanks!

PR #260, opened 2019-02-13
add zeta_zero

p15-git-acc:
This PR adds the `acb_dirichlet_zeta_zero` function. Functionally it should be almost the same as the mpmath implementation by Juan Arias de Reyna, but without providing the extra info about Gram/Rosser blocks, without the Illinois solver, without one of the heuristics for searching for zeros in an interval, and without hard-coding the first few exceptions to Rosser's rule.

The testing is kind of weak, because a different (and slow) code branch is used when `n` is large. The code is messy with multiple static functions in the file, and a custom data structure. Refinement of isolated zeros uses only bisection. In general the speed seems comparable or better than reported for mpmath.


fredrik-johansson:
Terrific!

First observation before I try to review this carefully: detailed comments (in zeta_zero.c) would be most welcome. I would like to see, for each helper function, a comment string giving a precise explanation of the inputs and outputs.

Regarding testing, it's definitely going to be hard to do strong testing in a randomized way. I propose testing against a table of known hard cases (Rosser exceptions, some zeros that are poorly separated).

fredrik-johansson:
Comments explaining each field of _zz_node_struct would also be useful.

I would be happy to work on the test code.

acb_dirichlet/zeta_zero.c
```diff
@@ -0,0 +1,748 @@
...
+    arb_init(t);
+    fmpz_init(k);
+
+    fmpz_set_si(k, n-2);
+    acb_dirichlet_gram_point(t, k, NULL, NULL, prec);
+    arf_set(&r->a, arb_midref(t));
+
+    fmpz_set_si(k, n-1);
+    acb_dirichlet_gram_point(t, k, NULL, NULL, prec);
+    arf_set(&r->b, arb_midref(t));
```
fredrik-johansson: Similarly, upper bound here.

acb_dirichlet/zeta_zero.c
```diff
@@ -0,0 +1,748 @@
...
+    p->prec = 0;
+    p->gram = NULL;
+    p->prev = NULL;
+    p->next = NULL;
+}
+
+static int
+zz_node_refine(zz_node_t p)
+{
+    p->prec = 2*FLINT_MAX(p->prec, 8);
```
fredrik-johansson: Here and elsewhere: starting with 8 bits seems inefficient for large n. I got roughly a 2x speedup for n from 10^6 to 10^12 by starting with 64 bits everywhere. It's possible that the initial precision should depend on the bit length of n -- but probably not everywhere, because the zeta function code also increases its precision internally. This doesn't need to be changed in this PR -- we can look at such optimizations later.

acb_dirichlet/zeta_zero.c
```diff
@@ -0,0 +1,748 @@
...
+    slong prec = 8;
+
+    arb_init(t);
+    arb_init(v);
+    acb_init(z);
+
+    arb_indeterminate(v);
+    while (arb_contains_zero(v))
+    {
+        prec *= 2;
```
fredrik-johansson: I want to move this to the end of the loop, for consistency with _definite_hardy_z.

acb_dirichlet/zeta_zero.c
```diff
@@ -0,0 +1,748 @@
...
+
+    return q;
+}
+
+static void
+split_interval(arb_t out,
+        const arf_t t1, const arb_t v1, slong sign1,
+        const arf_t t2, const arb_t v2, slong sign2, slong prec)
+{
+    if (sign1 == sign2)
```
fredrik-johansson: Explanation of what goes on here?

doc/source/acb_dirichlet.rst
```diff
@@ -636,3 +636,8 @@ Currently, these methods require *chi* to be a primitive character.
     multiplicities) of `\zeta(s)` in the region `0 < \mathcal{I}(s) \le T`.
     If at least *B* consecutive Gram blocks with union `[g_n, g_p)`
     satisfy Rosser's rule, then `N(g_n) \le n + 1` and `N(g_p) \ge p + 1`.
+
+.. function:: void acb_dirichlet_zeta_zero(acb_t res, const fmpz_t n, slong prec)
+
+    Sets *res* to the *n*-th nontrivial zero of `\zeta(s)`. Negative indices
+    give the conjugate zeros and `n = 0` is undefined.
```
fredrik-johansson: There should be a citation for https://ir.cwi.nl/pub/19724 and mpmath here.

fredrik-johansson:
I'm pleased that it seems to work correctly for the close zeros n = 8637740722917 and n = 8637740722918 :)

fredrik-johansson:
> The testing is kind of weak, because a different (and slow) code branch is used when `n` is large.

There's a way around this -- provide a version of zeta_zero that only uses the slow path even for small n; then we can test that separately.

acb_dirichlet/zeta_zero.c
```diff
@@ -0,0 +1,748 @@
+/*
+    Copyright (C) 2019 D.H.J. Polymath
```
fredrik-johansson: You are very welcome to share your real name (either here or on the credits page in the documentation) if you prefer. Well, it's up to you to be pseudonymous :)

Maybe also include

    Copyright (C) 2010 Juan Arias de Reyna

acb_dirichlet/zeta_zero.c
```diff
@@ -0,0 +1,748 @@
...
+    arb_t t;
+    fmpz_t k;
+    slong prec = 32;
+
+    arb_init(t);
+    fmpz_init(k);
+
+    fmpz_set_si(k, n-2);
+    acb_dirichlet_gram_point(t, k, NULL, NULL, prec);
+    arf_set(&r->a, arb_midref(t));
```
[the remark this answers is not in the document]
fredrik-johansson: Actually, even that is not sufficient to guarantee that we isolate the correct zero in the bisection... a hack workaround would be to check the accuracy of the balls for the Gram points...

p15-git-acc:
I split it into more files.

fredrik-johansson:
Can you explain why split_interval uses arb arithmetic, with the subsequent check

    !arb_contains_arf(t, &q->t) && !arb_contains_arf(t, &r->t)

Is there a reason for not just using arf arithmetic to compute the (tentative) new splitting point?

fredrik-johansson:
In general, this looks excellent. I will merge it and make some edits as I look through the code more thoroughly.

What are you doing next?

p15-git-acc:
> Is there a reason for not just using arf arithmetic to compute the (tentative) new splitting point?

The short reason is that arb arithmetic is more robust.

There are two code branches for computing the new splitting point. The opposite-sign branch just splits the interval in half, so arf arithmetic should be fine. The same-sign branch does something more clever which depends on the Z evaluations at the interval endpoints.

It would be possible to do the clever calculation with arf arithmetic and then verify that the new splitting point is between the two endpoints, but I'd be concerned that you would get splitting points that do not follow the intention of the clever formula and might lack whatever nice properties it might have. In the worst case it could lead to or exacerbate degenerate situations like the ones that mpmath works around with its extra heuristic that looks for zeros of the derivative of Z. That said, I don't have any real empirical or theoretical evidence that arf arithmetic wouldn't be enough.

p15-git-acc:
> What are you doing next?

Not only is the equal-signs splitting-point computed in a way that gives an arb interval when it could arguably use arf arithmetic instead, but the arb interval that it computes is unnecessarily inflated due to ball arithmetic reasons. I see a simple way to make the interval much tighter by rewriting the formula as a weighted arithmetic mean and computing the endpoints directly. So I'll plan to do that next.

p15-git-acc:
This is probably a dumb question, but can `arb_get_lbound_arf(t, x, prec)` ever give a negative `t` when `x` is nonnegative?

PR #262, opened 2019-02-15
tighten an interval, giving a speedup for large n

p15-git-acc:
zeta_zero is twice as fast for me now for large n (8637740722917 and 8637740722918).


fredrik-johansson:
Nice!

PR #266, opened 2019-02-18
add backlund_s

p15-git-acc:
Add `S(t)` as backlund_s.

### DIFF
--- a/acb_dirichlet/backlund_s.c
+++ b/acb_dirichlet/backlund_s.c
@@ -1,0 +1,71 @@
+/*
+    Copyright (C) 2019 D.H.J. Polymath
+
+    This file is part of Arb.
+
+    Arb is free software: you can redistribute it and/or modify it under
+    the terms of the GNU Lesser General Public License (LGPL) as published
+    by the Free Software Foundation; either version 2.1 of the License, or
+    (at your option) any later version.  See <http://www.gnu.org/licenses/>.
+*/
+
+#include "acb_dirichlet.h"
+
+void
+acb_dirichlet_backlund_s(arb_t res, const arb_t t, slong prec)
+{
+    mag_t m;
+
+    if (!arb_is_nonnegative(t))
+    {
+        arb_indeterminate(res);
+        return;
+    }
+
+    mag_init(m);
+
+    /*
+     * If the error radius of t is greater than 8/log(t) then use a fast bound.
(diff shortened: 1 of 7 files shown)


acb_dirichlet/backlund_s.c
```diff
@@ -0,0 +1,71 @@
...
+        arb_init(N);
+        acb_init(z);
+
+        /*
+         * The difference (N(t) - theta(t)/pi) loses precision from
+         * cancellation. Add extra bits to mitigate the loss.
+         */
+        acb_set_arb(z, t);
+        acb_dirichlet_hardy_theta(z, z, NULL, NULL, 1, prec);
+        prec += acb_bits(z);
```
fredrik-johansson: I also saw the use of arf_bits in the zeta zeros code, which is dubious. The number of bits doesn't necessarily correspond to the precision, accuracy or magnitude; in fact, it will fluctuate depending on the number of trailing zeros created by rounding. Shouldn't the precision just be doubled here?

fredrik-johansson: Alternatively, incremented by the log2 of the magnitude of t...

p15-git-acc: Thanks for the correction, I've changed both instances to now use log2 of magnitude.

p15-git-acc: There was a merge conflict that I've tried to resolve by undoing my changes to the zeta zeros code.

acb_dirichlet/isolate_hardy_z_zero.c
```diff
@@ -1218,8 +1218,6 @@ acb_dirichlet_zeta_nzeros(arb_t res, const arb_t t, slong prec)
         _arb_set_interval_fmpz(res, a, b);
     }
 
-    arb_set_round(res, res, prec);
```
fredrik-johansson: This should be put back in

acb_dirichlet/hardy_z_zero.c
```diff
@@ -148,6 +148,7 @@ void
 acb_dirichlet_hardy_z_zero(arb_t res, const fmpz_t n, slong prec)
 {
     arf_interval_t r, s;
+    mag_t m;
```
fredrik-johansson: The variable m can be removed.

PR #267, opened 2019-02-19
add zeta_zeros

p15-git-acc:
Adds zeta_zeros and hardy_z_zeros. Slightly refactors the hardy z zero function by separating out the refinement function. Tests and docs are included.

The theory behind these functions is that the single-zero versions have to isolate several zeros in a small range in order to get Turing's method to work. The multiple-zero versions use these already-isolated zeros when possible.

A nice performance test is to find the five zeros in a single interval, in the example in the presentation by Juan Arias de Reyna. The 'zeros' version is five times as fast as the 'zero' version.


acb_dirichlet.h
```diff
@@ -173,8 +173,11 @@ void _acb_dirichlet_isolate_gram_hardy_z_zero(arf_t a, arf_t b, const fmpz_t n);
 void _acb_dirichlet_isolate_rosser_hardy_z_zero(arf_t a, arf_t b, const fmpz_t n);
 void _acb_dirichlet_isolate_turing_hardy_z_zero(arf_t a, arf_t b, const fmpz_t n);
 void acb_dirichlet_isolate_hardy_z_zero(arf_t a, arf_t b, const fmpz_t n);
+void _acb_dirichlet_refine_hardy_z_zero(arb_t res, const arf_t a, const arf_t b, slong prec);
 void acb_dirichlet_hardy_z_zero(arb_t res, const fmpz_t n, slong prec);
+void acb_dirichlet_hardy_z_zeros(arb_ptr res, const fmpz_t n, slong len, slong prec);
 void acb_dirichlet_zeta_zero(acb_t res, const fmpz_t n, slong prec);
+void acb_dirichlet_zeta_zeros(acb_t res, const fmpz_t n, slong len, slong prec);
```
fredrik-johansson: res type should be acb_ptr

acb_dirichlet/zeta_zeros.c
```diff
@@ -0,0 +1,35 @@
...
+    the terms of the GNU Lesser General Public License (LGPL) as published
+    by the Free Software Foundation; either version 2.1 of the License, or
+    (at your option) any later version.  See <http://www.gnu.org/licenses/>.
+*/
+
+#include "acb_dirichlet.h"
+
+void
+acb_dirichlet_zeta_zeros(acb_ptr res, const fmpz_t n, slong len, slong prec)
+{
```
fredrik-johansson: Return if len <= 0

acb_dirichlet/isolate_hardy_z_zero.c
```diff
@@ -1177,6 +1177,108 @@ _acb_dirichlet_exact_zeta_nzeros(fmpz_t res, const arf_t t)
...
+
+    return i;
+}
+
+void
+acb_dirichlet_hardy_z_zeros(arb_ptr res, const fmpz_t n, slong len, slong prec)
+{
+    slong r; /* the number of zeros added in one pass */
+    slong c; /* the total number of zeros added so far */
+    fmpz_t k;
```
fredrik-johansson: Return if len <= 0

fredrik-johansson:
I think zero(n) should be implemented by calling zeros() with a length of 1 (it must be assumed that zeros() with a length of 1 does the correct and optimal thing), rather than having duplicated code for this case.

p15-git-acc:
Thanks, some stuff in the latest patch, but there are still two potential issues I could work on:

* zeta_zeros currently doesn't allow ranges that include negative numbers or zero. I could fix this so that it's allowed, including ranges that include both negative and positive numbers.
* There is still duplication between the code to isolate a single hardy z zero and the code to compute a range of hardy z zeros.

fredrik-johansson:
Please do!

fredrik-johansson:
Though, I'm not sure if it makes more sense to skip index n = 0 or define zero(0) = indeterminate. Perhaps the latter?

fredrik-johansson:
Also negative indices may as well be allowed for the Z-function zeros.

p15-git-acc:
After thinking about it more, I might just get rid of all n < 1 indices of zeros. I feel like it was added to mpmath because it was cute and could be done in just a couple lines of code, but for ranges it's way more annoying.

fredrik-johansson:
Fair enough.

p15-git-acc:
Added a static helper function that isolates multiple zeros, filling an arf_interval array.

PR #269, opened 2019-02-21
a multi riemann zeta platt example to discuss

p15-git-acc:


p15-git-acc:
The code is an interpretation of part of Platt's "Isolating some non-trivial zeros of zeta":
https://www.researchgate.net/publication/305338841_Isolating_some_non-trivial_zeros_of_zeta
which is based on Platt's thesis:
https://people.maths.bris.ac.uk/~madjp/thesis5.pdf
where the part relevant to this PR is a windowed variant of Booker's "Artin's conjecture, Turing's method and the Riemann hypothesis":
https://arxiv.org/pdf/math/0507502.pdf

fredrik-johansson:
Could you provide some explanation of what is still missing / wrong / in need of optimization?

p15-git-acc:
General difficulties:
- Individual components of the algorithm are not particularly worth including into arb in isolation.
- The only references for the windowed variant of Booker's algorithm are Platt's thesis and a separate publication of the relevant sections of the thesis. This means that the inevitable typos or errors haven't been ironed out yet, and any gaps or missing details have not yet been filled in. There are reference implementations for related algorithms, but I don't think there's a reference implementation for this algorithm.
- There are like a dozen cumulative error bounds which are probably dominated by only a couple. This means that math or programming errors in the dominated bounds will not be caught in normal tests.

Missing features in this PR:
- Grid upsampling by padding the F(x) vector with zeros is not implemented.
- Whittaker-Shannon upsampling at arbitrary points (e.g. Gram points) is not implemented.
- The parameters A, B, h, J, K, sigma, and prec are not automatically selected.
- It is not integrated with the zeta zero isolation/refinement code.
- Constraints on the parameters of the bound calculations are not enforced.
- Comments are missing in the code.
- Missing .rst docs.
- Missing tests.

I'm least confident about the part of the code near the triple sum related to Taylor's theorem, both in terms of the correctness of the approximation and its error bounds. This is also near where I'm using the padded DFT's that have been suggested to be replaced by polynomial multiplication. The references are not completely explicit about the indexing of the sums, and I remember having some question in the De Bruijn Newman thread about whether the constraints on the parameters of the error bounds are met.

p15-git-acc:
I'm having trouble understand the very last proof in https://www.researchgate.net/publication/305338841_Isolating_some_non-trivial_zeros_of_zeta which is the proof of Lemma C.3. Could someone explain just the first line of that proof, or help find a typo or error if it has one?

This Lemma gives one of the two error bounds for Whittaker-Shannon upsampling.

simonpuchert:
From a quick glance, it looks like the condition in the sum should be changed from
|n - t_0 / A| > N_s
to
|n - t_0 * A| > N_s.

p15-git-acc:
@simonpuchert Thanks! That's so close but after that fix it's still off by a factor of A inside the sinc. Maybe it's a typo that you've spotted and then also a mistake in the proof on top of the typo?

fredrik-johansson:
I have checked at least superficially that the code agrees with the formulas in Platt's paper.

p15-git-acc:
@fredrik-johansson thanks for checking it! I'm still hung up on the Lemma C.3 issue. @djplatt?

fredrik-johansson:
Dave clarified this by email:

> On Lemma C.3 as it stands, I fear you are right and the condition under the summation should read |n-t_0 A| > N_s. This also means that the first sinc in the proof should be sinc(N_s) not sinc(N_s/A) so I have over stated the error by a factor of A. Apologies. Needless to say, my code implements to summation as I meant, rather than as I said, and since A>>1 the error bound is actually tighter than I claimed.

fredrik-johansson:
Dave Platt also told me that je's actively working on the generic L-functions package for the LMFDB, and that it will be better than the "Isolating some non-trivial zeros of zeta" algorithm even in the special case of zeta. Still waiting for more info about that...

djplatt:
Hi Fredrik


What I tried to say was that the error bounds we (mostly Min Lee and Andy) have produced for the generic case may even beat the specific ones I came up with for zeta in my thesis. I'd have to check. There is no way our generic code will ever beat the zeta specific version overall, but incorporating any improved bounds might improve the latter slightly.


As far as I know, the fastest rigorous procedure for zeta at height remains the windowed version of Andy's "Artin..." method. I did have some discussions with Jan Buthe a while back about swapping the Gaussian kernel for something else. He is somewhat of an evangelist for Logan's function (see for example his zeros paper) but I never got around to trying it and it would mean redoing all the error bounds, again.


Sorry if I confused the issue.


Best


Dave


________________________________

fredrik-johansson:
@djplatt Thanks again for clarifying!

p15-git-acc:
I don't understand all the details of how Turing's method is used here. In particular I don't understand how N(t1) is computed in the discussion following Theorem 4.2 in the "Isolating some non-trivial zeros of zeta" paper. It says "Note that even though we do not know N(t1) a priori, we know it to be an integer and Theorem 4.2 will allow us to determine which one." I went back to the thesis and it says "If the assumed number of zeros up to T is incorrect, it will quickly become apparent through a contradiction of Theorem 2.5.2 (or one of its later improvements). This process is made absolutely rigorous in [Booker's "Artin..." paper]." In Booker's paper the description of Turing's method in section 4 is a little complicated because it works out the details necessary to apply it to an arbitrary L-function and it uses different notation.

One thing in particular that I don't understand is that Booker's method appears to require some information about the zeros on both sides of each endpoint of the interval in which the zeros are to be isolated rigorously, whereas Platt's method appears to not ask for any information below t1.

djplatt:
Hi Fredrick


1) The generic code we have been working on is aimed at producing the data required to populate the LMFDB. One specifies the degree r, the shifts mu_i that appear in the gamma factors, the conductor and as many Euler polynomials as are required for the target accuracy. It then computes L on a lattice, aiming to get accurate results to height 64/r and isolates the zeros to that height. It uses Buthe's method to verify that all the zeros have been found and, for rank 0 or 1 L-functions, this will be a rigorous validation of RH. Beyond rank 1, the code finds the first definitely non-zero taylor coefficient and takes this to be the rank. It is of course possible that an earlier coefficient was in fact a very small number pretending to be zero, hence no rigour, unless the user has other reasons for knowing the rank. It supports computing "special values" off the critical line but accuarcy degrades rapidly as one moves out. We envisage using this to compute L(1) etc.


2) When validating the list of zeros for zeta, my code will, in the vast majority of cases, already know the zero count N(t1) at the start of the region since it is the zero count to the end of the previous region. If it is the first region in a run, or if the RH verification failed for the previous region, then we do have to compute N(t1) from scratch and this requires that we have some knowledge about the zeros just below t1. I could scribble a quick note on how to translate Andy's generic version for zeta if that would help.


Best regards


Dave


________________________________

p15-git-acc:
Hi github issue followers,

Here's a proposed rough roadmap for an implementation in arb of fast calculation of multiple consecutive nontrivial zeros of zeta to medium precision, near height t0:
 1) Given t0, somehow automatically set the ~7 parameters for Platt's algorithm.
 2) Evaluate Platt's f(t) on a lattice near t0 using the code in this PR. Keep only the middle ~1/2(?) of the points, because windowing causes the others to have low prec.
 3) Use Shannon-Whittaker interpolation of the lattice to certify some Rosser/Gram blocks as 'good'. This works only near the middle of the points kept in the previous step. It doesn't work near the ends because the interpolation needs several lattice points above and below the target point.
 4) Use Arias de Reyna's variant of Turing's method with Trudgian's bound to determine N(t) when t is the center Gram point of a long enough contiguous stretch of 'good' blocks. This step also loses information from the ends because it needs padding blocks on both sides, and this step will abort if fewer than two points with known N(t) are found.
 5) Use Arias de Reyna's subdivision strategy, but with Shannon-Whittaker interpolation instead of direct evaluation of Z(t), to isolate all zeros between pairs of points where N(t) is known. This step would hang or abort if the isolation is not possible due to the lattice (low point density, low lattice precision, or proximity to the ends of the lattice).
 6) Use bisection and Shannon-Whittaker interpolation to refine isolated zeros. Precision is limited not only by the number of bisection steps but also by the lattice. I guess it could be possible to use automatic differentiation to differentiate through the Shannon-Whittaker interpolation if this would be helpful for a better refinement algorithm, but it's complicated by discontinuities from the sum limit in Lemma C.3 and I don't think derivative error bounds would be available.

As you can see, this would kind of give up on the Turing algorithm variant described by Booker, but my rationalization is that Arias de Reyna's variant shouldn't waste too many zeros and that because several Shannon-Whittaker interpolation calls are required per zero for refinement then it should be OK to use a few calls for isolation as well.

This plan would rely heavily on Shannon-Whittaker interpolation of a function that is re-windowed from `h` to `H` in the definition of W(t) in "Isolating some nontrivial zeros of zeta". I have an intuitive understanding of the `h` that is used for windowed DFT/IDFT, but I don't really understand the `H` that is used in concert with Shannon-Whittaker interpolation. The paper says "When up-sampling, we set H = 2089.0/16384.0 and used 70 points either side of t0 using every 5th one in the sum." where I assume t0 in that context is the height of the point whose value is to be computed by Shannon-Whittaker interpolation. I don't really understand why to use every 5th point, or whether it's every 5th point before or after their 32x upsampling-by-nearly-zero-frequency-padding step, or how taking every 5th point would affect the error bound (would it require replacing `A` with `A/5` in the bound?). I also don't know whether they re-centered the H window on the height to be evaluated by interpolation each time the interpolation function is called, or if the same H windowed W(t) function is used for all interpolation calls for the same lattice. In the thesis, the choice of `H` in the implementation is not mentioned at all except to say that "the Gaussian width H >0 need not be and indeed will not be the same as the h of Equation 5.1.1.".

By the way, there's also another Lemma C.3 typo. A lower case `h` is mentioned only one time in the lemma and it should have been `H`.

simonpuchert:
I found a small possible improvement:

You can drop a factor of `2^beta` from `Z` in the bound of Lemma C.3.
The current proof uses
`(t0 + n/A)^beta <= 2^beta t0^beta + 2^beta (n/A)^beta * 1[n/A >= t0].`
As `beta < 1`, this can be improved to
`(t0 + n/A)^beta < 2^beta t0^beta + (n/A)^beta * 1[n/A >= t0]`
(generally, the factor would be `max(2^(beta-1),1)`), which isn't a lot better, but since it makes the code simpler, it's still somewhat useful.

p15-git-acc:
@simonpuchert I don't understand the improvement completely enough to change the code but here are my notes as I was trying to work through it.

If you have positive `a, b, c` then `(a+b)^c <= (2a)^c + (2b)^c = 2^c * a^c + 2^c * b^c` which can be improved to `(a+b)^c <= 2^c * a^c + 2^c * b^c * 1[a <= b]`. The inequality can be made even tighter by refining the second `2^c` term, so `(a+b)^c <= 2^c * a^c + max(1, 2^(c-1)) * b^c * 1[a <= b]`. In the last section of the proof of C.3 the integral is split into two parts where the integration limits on the second part somehow corresponds to the conditional in the bound.

It would be nice to get rid of `Z` completely. I don't really understand your explanation, but when you refer to the bound for the `sinc` function I guess you mean `|sin(x)/x| <= |1/x|`.

simonpuchert:
The idea is that (with `n = k + t_0*A` and `k >= N_s`)
`|W(n / A) sinc(n - t_0 * A)| <= 3 (t_0 + k / A)^beta exp(-k² / (2*A²*H²)) * 1/(π k)`, which is a decreasing function times the exponential term (the difference to Platt's proof up to here is that he used the bound `|sinc(n-t_0*A)| <= 1/(π N_s)` instead, so the "rest" is not a decreasing function), thus

`|W(n / A) sinc(n - t_0 * A)| <= 3/(π N_s) * (t_0 + N_s / A)^beta * exp(-k^2 / (2*A^2*H^2)).`

and then we could replace the term `(t_0 + N_s / A)^beta` with `(2t_0)^beta` to fit the formula for `Y`, but since we have to calculate `(t_0 + N_s / A)^beta` anyway because of the term `X`, we can just leave that step out, reducing the bound to
`6/(π N_s) * (X + 1/sqrt(2) (t_0 + N_s / A)^beta * A * H * Gamma(1/2, (N_s)^2 / (2*A^2*H^2)))`.

So in summary, you can remove `Z` completely and (if you want) you can reduce the factor in `Y` (and since the same factor also appears in `X`, maybe factor it out).
Don't expect the bounds to be a lot better though, as `Z` is very small for `t_0 >> N_s / A`, which is presumably the most interesting case.

ValZapod:
Maybe somebody (author?) will fix Lemma C.3 in the paper? Mmm?

fredrik-johansson:
Closing this as obsolete.

PR #270, opened 2019-02-22
Add functions specific to Gram points

p15-git-acc:
These are twice as fast and return integers rather than arb intervals. Also the code is simpler and can be used for testing against the more general case.


fredrik-johansson:
Seems OK.

PR #275, opened 2019-04-19
platt whittaker shannon interpolation

p15-git-acc:
This PR implements the Whittaker-Shannon interpolation component of Platt's method of isolating zeta zeros. Here's an example plot of the interpolation (with error bounds):
interpolation plot

### DIFF
--- a/acb_dirichlet.h
+++ b/acb_dirichlet.h
@@ -195,6 +195,76 @@ acb_dirichlet_zeta_zero(acb_t res, const fmpz_t n, slong prec)
     acb_dirichlet_zeta_zeros(res, n, 1, prec);
 }
 
+/* Platt zeta zeros */
+
+typedef struct
+{
+    slong len;
+    arb_ptr p;
+    arb_struct Xa;
+    arb_struct Xb;
+}
+acb_dirichlet_platt_c_precomp_struct;
+typedef acb_dirichlet_platt_c_precomp_struct acb_dirichlet_platt_c_precomp_t[1];
+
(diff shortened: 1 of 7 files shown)


p15-git-acc:
I think all of the issues flagged by 'codacy' are false positives.

fredrik-johansson:
> I think all of the issues flagged by 'codacy' are false positives.

Yes, it looks OK to me.

fredrik-johansson:
Apologies for the delay -- the code looks great.

p15-git-acc:
My progress on implementing the platt zeta zeros algorithm (in particular selecting tuning parameter values automatically) got hung up on unexpectedly inefficient incomplete gamma function evaluation. I found https://arxiv.org/abs/1803.07841 using google, which looked promising for the transition region where the trouble appeared to be, but I don't know if it's suitable for implementing in arb.

fredrik-johansson:
Inefficient or inaccurate? It doesn't seem like it should be the bottleneck in the algorithm.

If it's just used for estimates and doesn't require bounds, that would make things easier.

p15-git-acc:
> Inefficient or inaccurate?

Imprecise but becomes inefficient when working around the imprecision.

> It doesn't seem like it should be the bottleneck in the algorithm.

For most parameter combinations it's not the bottleneck in the algorithm in this pull request (which itself will probably not be the bottleneck in the platt zeta zeros algorithm), but it's the bottleneck in a script that I tried to use to help develop tuning parameter heuristics.

PR #279, opened 2019-06-08
platt multieval

p15-git-acc:


fredrik-johansson:
Is this ready to merge?

p15-git-acc:
Yes I intended for it to be mergeable when I submitted the PR. Still I would be happy to make changes if you see any way of improving the code or the design.

fredrik-johansson:
Sure, just asking since you updated the incomplete gamma function code afterwards. Will have a closer look soon.

p15-git-acc:
Now that you mention it, there are a couple things. One could be to tighten the interface by changing several `slong` to `ulong` for nonnegative integer arguments in the new functions. Another could be to improve the testing. Currently multieval is not tested with realistic parameter values because I'd thought the realistic ones would be too slow, but there are a couple things I could do. First I could try to find the smallest T where the multieval gives informative output and only do the test for the best tuning parameter values for this T, and second I could do the comparison against single eval by only checking a random subset of the multieval points. Checking every point is likely too slow, and the multieval algorithm gives uninformative output if A or B is too small.

p15-git-acc:
I added the extra test, but I'm leaning against changing more slong to ulong because C doesn't really enforce signed/unsigned restrictions.

PR #287, opened 2019-08-07
heuristic tuning parameters for platt hardy z zeros

p15-git-acc:
Heuristically picks ten tuning parameter values for Platt's method of finding zeros of the Hardy Z function.


fredrik-johansson:
Why the requirement n >= 10^4?

p15-git-acc:
It could be lowered somewhat if necessary. The reason the lower limit exists is that the method has to evaluate the function on a grid and only a fraction of the points in the middle of the grid end up having enough precision to be usable. For example according to Platt's thesis, "Below t = 5000 our algorithm is not effective so the zeros below this height were located using mpmath."

Issue #285, opened 2019-07-25
add approximate derivative of gaussian windowed whittaker shannon interpolation for platt zeta zeros


p15-git-acc:
This would be for platt zeta zeros refinement; when the root finder detects an f(x) interval containing zero, an approximate f'(x) can help guess eps so that f(x-eps) and f(x+eps) might have definite and opposite signs.

PR #290, opened 2019-08-12
interpolation derivative for zeros refinement under limited precision

p15-git-acc:
closes https://github.com/fredrik-johansson/arb/issues/285


p15-git-acc:
Codacy is flagging an uninitialized precomputation variable that was a false positive last time and I think it's a false positive this time as well.

p15-git-acc:
I'm not totally sure that all of the `len` and `hlen` lengths are correct everywhere for the automatic differentiation but the results appear to agree with finite differences approximations. The new sinc_pi series could be moved into its own file. Would that be worth doing in this PR? I copied it with minimal modifications from the sinc series function that is already in arb.

p15-git-acc:
Moved the sinc_pi automatic differentiation to its own file plus test and doc, copied from sinc with minimal modifications.

PR #291, opened 2019-08-16
add csc_pi for arb_t and acb_t types

p15-git-acc:
I was looking in the gamma.c files to try to improve the upper incomplete gamma function and I found a couple todo lines requesting `csc_pi`.

### DIFF
--- a/acb.h
+++ b/acb.h
@@ -723,6 +723,13 @@ void acb_sin_cos_pi(acb_t s, acb_t c, const acb_t z, slong prec);
 void acb_tan_pi(acb_t r, const acb_t z, slong prec);
 void acb_cot_pi(acb_t r, const acb_t z, slong prec);
 
+ACB_INLINE void
+acb_csc_pi(acb_t y, const acb_t x, slong prec)
+{
+    acb_sin_pi(y, x, prec);
+    acb_inv(y, y, prec);
+}
+
 void acb_sinc(acb_t res, const acb_t z, slong prec);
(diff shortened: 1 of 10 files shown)


fredrik-johansson:
acb_csc_pi should avoid dividing by something exponentially large away from the real line axis. See acb_csch.

p15-git-acc:
After trying this, it gave correct but not usually better intervals. I guess I don't understand why it would be bad to create an output that has a large negative exponent by taking the reciprocal of an intermediate calculation that has a large positive exponent. I could understand if the change would avoid an interval 'dependency problem' or avoid a numerical cancellation or if large exponents (in absolute value) could be avoided altogether but I don't see how those apply here.

For reference, I copied acb_csch replacing acb_exp with acb_exp_pi_i and making other small changes (like reversing inequalities in some conditionals, switching some realref and imagref, and multiplying the final result by i) with the idea of keeping exponentials in the intermediate calculations always small.

fredrik-johansson:
Take a wide complex interval: z = [1000 +/- 10]*I, say. Then 1/sin(z) ends up dividing by zero.

p15-git-acc:
OK fixed. There was a bug in my check of which version was better, where it preferred the old way when both the old and new way gave nan +/- inf, which happened a lot. After fixing this, I see that the way that you suggest usually has better intervals and gives nan +/- inf much less often.

fredrik-johansson:
Thanks!

Issue #278, opened 2019-06-05
Function _hypot and Microsoft Windows API function duplicate name


JackBoosY:
Hi, developer.
Recently I want to update the arb port on vcpkg to version 2.16.0, but I encountered an error while building:
> F:\vcpkg\buildtrees\arb\src\2.16.0-f475b1df04\acb_dirichlet\stieltjes.c(343): error C2491: '_hypot': definition of dllimport function not allowed

So I looked at the _hypot statement in the code:
`static double _hypot(double x, double y) { return sqrt(x * x + y * y); }`

This error is too strange, so I searched for the function name on google and found it to be a windows API name ([document](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/hypot-hypotf-hypotl-hypot-hypotf-hypotl?view=vs-2019)).

So I had to change the name of this function to fix this problem([#6763](https://github.com/microsoft/vcpkg/pull/6763)).
Here is my patch: [fix-build-error.zip](https://github.com/fredrik-johansson/arb/files/3255354/fix-build-error.zip)

At last, I suggest replacing this function name.

Thanks.

fredrik-johansson:
Fixed. Thanks!

saraedum:
This now clashes on my system:
```
stieltjes.c:343:15: error: static declaration of '__hypot' follows non-static declaration
 static double __hypot(double x, double y) { return sqrt(x * x + y * y); }
               ^~~~~~~
In file included from /home/jule/proj/conda/miniconda3/envs/flatsurf/x86_64-conda_cos6-linux-gnu/sysroot/usr/include/features.h:356:0,
                 from /home/jule/proj/conda/miniconda3/envs/flatsurf/x86_64-conda_cos6-linux-gnu/sysroot/usr/include/math.h:28,
                 from stieltjes.c:12:
/home/jule/proj/conda/miniconda3/envs/flatsurf/x86_64-conda_cos6-linux-gnu/sysroot/usr/include/bits/mathcalls.h:163:1: note: previous declaration of '__hypot' was here
 __MATHCALL (hypot,, (_Mdouble_ __x, _Mdouble_ __y));
```

saraedum:
Section 7.1.3 of C99 says:

> —All identifiers that begin with an underscore and either an uppercase letter or anotherunderscore are always reserved for anyuse.
> —All identifiers that begin with an underscore are always reserved for use as identifierswith file scope in both the ordinary and tag name spaces.

[...]

PR #292, opened 2019-09-11
Fix clash of __hypot

saraedum:
see #278, __ are reserved according to the C99 standard.


fredrik-johansson:
`arb_hypot` is the name of a different function.

How about `_d_approx_hypot` or whatever :)

saraedum:
Sure :)

PR #304, opened 2020-04-12
fix doc typo

Joel-Dahne:
Trying to do some automatic parsing of the documentation in relation to [Arblib.jl](https://github.com/kalmarek/Arblib.jl) I ran into this parsing error.


fredrik-johansson:
Thanks!

PR #308, opened 2020-05-04
Remove long

Joel-Dahne:
In general arb uses `slong` for 64 bits signed integers, in a few places `long` was used instead. I believe all of these cases are just typos and this pull request fixes those. There seems to be no other occurrences as determined by `grep -R " long"` (notice the space before `long`).


fredrik-johansson:
Thanks!

PR #322, opened 2020-09-02
add fransen robinson constant as an integral example

p15-git-acc:
While looking at integrals of gamma-related functions with the hope of improving the incomplete gamma function implementation for zeta zeros parameter tuning I noticed this constant on Wikipedia so I've tried adding it as an integral example. Later I saw on its oeis entry that you've already written this code lol, but I thought I'd submit this pr anyway.


fredrik-johansson:
Sure, why not :-)

PR #323, opened 2020-09-04
avoid asymptotic expansion for the upper incomplete gamma function under some conditions

p15-git-acc:
This pull request improves the implementation of `acb_hypgeom_gamma_upper` by no longer dispatching to `acb_hypgeom_gamma_upper_asymp` under certain conditions when doing so can be predetermined to give an uninformative result (nan +/- inf). The improvement is particularly noticeable when `s` is large compared to `z` and to the number of bits of working precision.

The shaded region of this [figure](https://dlmf.nist.gov/13.7#F1) shows where `z` is too close to the origin for the asymptotic expansion to work, and the scale of that region depends on `s`. The old code uses the asymptotic expansion when `z` is outside a disk whose radius is proportional to the working precision, without taking `s` into account. The new code adds a condition requiring `z` to also be outside of the shaded region.

This PR does not address the transition region problems when `z` is near the boundary of the shaded region.

There was a typo on line https://github.com/fredrik-johansson/arb/blob/master/acb_hypgeom/u_asymp.c#L73 where `if (R == 0)` should be `if (*R == 0)`. It looked scary but I'm pretty sure its effects were masked.

### DIFF
--- a/acb_hypgeom/gamma_upper.c
+++ b/acb_hypgeom/gamma_upper.c
@@ -11,6 +11,10 @@
 
 #include "acb_hypgeom.h"
 
+int
+acb_hypgeom_u_asymp_determine_region(const mag_t r,
+    const mag_t zlo, const acb_t z);
+
 void
 acb_hypgeom_gamma_upper_asymp(acb_t res, const acb_t s,
         const acb_t z, int regularized, slong prec)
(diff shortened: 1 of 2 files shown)


fredrik-johansson:
Looks reasonable, thanks.

PR #324, opened 2020-09-05
fix a zeta zeros issue

p15-git-acc:
Fixes a mistake in the implementation of Platt's algorithm for finding multiple zeta zeros at once. The bug meant that the code was checking a regularity condition that was different than the intended condition. I don't know about the practical impact. It also caused the code to unnecessarily abort in certain circumstances when no results could be found.


fredrik-johansson:
Nice :-)

Issue #276, opened 2019-04-28
arb_hypgeom_gamma_upper(357, 356) gives nan +/- inf


p15-git-acc:
According to Pari/GP and Mathematica it should be
1.2397102993749311592885574337209916963 E755

```c
#include "arb_hypgeom.h"

int main()
{
    slong prec = 300;
    arb_t g, s, x;
    arb_init(g);
    arb_init(s);
    arb_init(x);
    arb_set_ui(s, 357);
    arb_set_ui(x, 356);
    arb_hypgeom_gamma_upper(g, s, x, 0, prec);
    arb_printd(g, 30);
    flint_printf("\n");
    arb_clear(g);
    arb_clear(s);
    arb_clear(x);
    flint_cleanup();
    return 0;
}
```

Edit: It gives the expected result if `prec` is increased from 300 to 400. This is probably because of conditionals in `acb_hypgeom_u_asymp`. For my application I think I need only a few bits of precision in the output, not anywhere near 300 or 400 bits.

p15-git-acc:
After some more testing I think the problem is in the determination of whether or not the asymptotic expansion should be used. Here's an example with more generic input (not exact or integers or tiny intervals around integers):

```c
#include "arb_hypgeom.h"
#include "acb_hypgeom.h"

static void
check_incomplete_gamma(const char *sstr, const char *xstr, slong prec)
{
    arb_t g, s, x;
    acb_t gc, sc, xc;

    arb_init(g);
    arb_init(s);
    arb_init(x);
    acb_init(gc);
    acb_init(sc);
    acb_init(xc);

    arb_set_str(s, sstr, prec);
    arb_set_str(x, xstr, prec);
    arb_add_error_2exp_si(s, -200);
    arb_add_error_2exp_si(x, -200);
    acb_set_arb(sc, s);
    acb_set_arb(xc, x);
    flint_printf("s: "); arb_printd(s, 20); flint_printf("\n");
    flint_printf("x: "); arb_printd(x, 20); flint_printf("\n");
    arb_hypgeom_gamma_upper(g, s, x, 0, prec);
    flint_printf("gamma_upper(s, x): ");
    arb_printd(g, 20); flint_printf("\n");
    acb_hypgeom_gamma_upper_1f1b(gc, sc, xc, 0, prec);
    flint_printf("gamma_upper_1f1b(s, x): ");
    acb_printd(gc, 20); flint_printf("\n");
    flint_printf("\n");

    arb_clear(g);
[...]

Issue #166, opened 2016-09-11
incomplete gamma inaccurate in asymptotic range


pascalmolin:
Currently, Gamma(0, 110) returns a 10^-4 ball (test code below).

A precise result is needed for L functions (see commit acc874e from #165)
I guess one needs asymptotic expansion like
http://dlmf.nist.gov/8.11#i

example code

```
#include "acb_hypgeom.h"

int main()
{
    slong prec = 165;
    acb_t g, s, x;
    acb_init(g);
    acb_init(s);
    acb_init(x);
    /* s = 0 */
    acb_zero(s);
    /* x = 110 */
    acb_set_si(x, 110);
    acb_hypgeom_gamma_upper(g, s, x, 2, prec);
    acb_printd(g, 30);
    flint_printf("\n");
    acb_clear(g);
    acb_clear(s);
    acb_clear(x);
    flint_cleanup();
    return EXIT_SUCCESS;
}
```

fredrik-johansson:
The asymptotic 2F0 series is used when |z| is large enough, but the cutoff is not optimized. Right now, it uses the 1F1 series when the 2F0 is estimated to give less than prec bits of accuracy, even if the accuracy with the 1F1 series is much less than that (because of cancellation). The algorithm should really estimate which of the 1F1 and 2F0 formulas gives the most accuracy; something similar is done in the 1F1 code right now.

However, even with a more intelligent cutoff, there will be some loss of accuracy in the intermediate range.

For L-functions, we don't need to improve the incomplete gamma function in general; it's enough to make it more precise in this special evaluation mode.

It's very easy if s is exact: the L-functions just needs to wrap acb_hypgeom_gamma_upper in a loop that increases the working precision used to call this function until the error is less than 2^-prec (z is not exact here, but since it's just a rational multiple of pi, we can recompute it to the working precision).

If s is not exact, we could evaluate Gamma(s,z) at the midpoint and bound the propagated error.

I think we should get a good bound on |Gamma(S,z)| on a complex ball S surrounding s (to use the Cauchy integral formula) from \int_z^{\infty} |t^{S-1} e^{-t}| dt, or directly for the s-derivative from \int_z^{\infty} |t^{s-1} log(t) e^{-t}| dt.

fredrik-johansson:
There is now a trivial solution with the new integration code.

What is less trivial is to select algorithms automatically.

PR #326, opened 2020-09-07
improve upper incomplete gamma algorithm selection for nonnegative real s and z

p15-git-acc:
When s and z are nonnegative real, the asymptotic expansion is used when `(z-c)^8 > (a*s)^8 + (b*prec)^8` where `a=1.029287542, b=0.3319411658, c=2.391097143`. For reference the `acb_hypgeom_u_use_asymp` condition has `a=0, b=0.69314718055994530942, c=0` in this notation.

Closes https://github.com/fredrik-johansson/arb/issues/276.
Closes https://github.com/fredrik-johansson/arb/issues/166.

The [magic constants](https://en.wikipedia.org/wiki/Magic_number_(programming)) were found by locating the best transition point for each point in the grid prec=16..400 s=0.5..200.5 then contour plotting the results, guessing a nonlinear model, and fitting the constants. The mean squared error for the predicted x transition point as a function of s and prec is about 1. The model looks solid enough to me that extrapolating beyond the training grid seems OK.

Machine floats are used for speed, and I'm not sure if overflow would be a problem. This PR only deals with nonnegative real s and z, so there are still algorithm selection problems when s and z have imaginary parts. I'm not totally sure what's going on with integer values of s. Surprisingly I haven't noticed any actual transition region issues, in other words I haven't noticed a point where neither the 1f1 nor the asymp algorithms were good enough, but I haven't looked for this specifically.

### DIFF
--- a/acb_dirichlet/platt_c_bound.c
+++ b/acb_dirichlet/platt_c_bound.c
@@ -12,33 +12,6 @@
 #include "acb_dirichlet.h"
 #include "arb_hypgeom.h"
 
-/* Increase precision adaptively. */
-static void
-_gamma_upper_workaround(arb_t res, const arb_t s, const arb_t z,
-        int regularized, slong prec)
-{
-    if (!arb_is_finite(s) || !arb_is_finite(z))
-    {
-        arb_indeterminate(res);
-    }
-    else
-    {
-        arb_t x;
-        slong i;
-        arb_init(x);
-        for (i = 0; i < 5; i++)
-        {
-            arb_hypgeom_gamma_upper(x, s, z, regularized, prec << i);
-            if (arb_rel_accuracy_bits(x) > 1)
-            {
-                break;
-            }
-        }
-        arb_swap(res, x);
-        arb_clear(x);
-    }
-}
-
 static void
 _arb_pow_si(arb_t res, const arb_t x, slong y, slong prec)
 {
@@ -175,7 +148,7 @@ _pre_c_p(arb_ptr res, slong sigma, const arb_t h, ulong k, slong prec)
 
         arb_set_si(x, k + l + 1);
         arb_mul_2exp_si(x, x, -1);
-        _gamma_upper_workaround(x, x, u, 0, prec);
+        arb_hypgeom_gamma_upper(x, x, u, 0, prec);
         arb_mul(res + l, res + l, x, prec);
     }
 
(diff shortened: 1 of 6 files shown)


p15-git-acc:
I think the Sage CI failed for reasons that aren't my fault.

p15-git-acc:
I'm going to close this, make some changes, and reopen. Maybe it will force Sage CI to try again.

p15-git-acc:
The last commit added some accuracy regression tests, removed the wrapped incomplete upper gamma function from zeta zeros code, and improved the algorithm selection. Instead of checking the 8-norm directly, it first checks the 1-norm and infinity-norm. This should be faster in most cases and is also robust to overflow in the 8-norm; if overflow somehow causes the wrong algorithm to be selected then incrementing the precision eventually causes the other checks to kick in.

fredrik-johansson:
It seems fine apart from the potential overflow issue with doubles. I want to be sure that it goes into acb_hypgeom_gamma_upper_asymp when given something like z = 1e10000, s = 1 (and not just by accident).

fredrik-johansson:
> I think the Sage CI failed for reasons that aren't my fault.

Yes, don't worry about that.

p15-git-acc:
OK it now uses mag_t instead of machine float. I added and tested an internal helper function to efficiently compare a^n to b^n + c^n.

p15-git-acc:
The helper function isn't rigorous, because it doesn't use intervals and doesn't pay attention to mag rounding. It doesn't matter for what I'm using it for, but maybe it could be changed to one or two one-sided rigorous functions like `mag_lt_norm_ui`/`mag_gt_norm_ui` that return 0/1 instead of the current non-rigorous two-sided function that returns -1/0/1. I'll close this PR for now and try that.

fredrik-johansson:
I think the previous version using doubles is fine. You just need to add an initial check for whether z is super-huge, and in that case skip the double computation.

fredrik-johansson:
Cf https://github.com/fredrik-johansson/arb/blob/master/acb_hypgeom/erfc.c

p15-git-acc:
What if s is also super-huge though

fredrik-johansson:
You could check that in the same way.

p15-git-acc:
If x and s are both huge then whether or not to use the asymptotic expansion depends on their relative sizes.

I've reopened this with a `_mag_gt_norm_ui` function that should do the one-sided comparison with correct rounding.

fredrik-johansson:
> If x and s are both huge then whether or not to use the asymptotic expansion depends on their relative sizes.

Correct, but if they are both huge then the non-asymptotic expansion will fail anyway.

> I've reopened this with a _mag_gt_norm_ui function that should do the one-sided comparison with correct rounding.

That's fine. I will have a look.

fredrik-johansson:
Ok, merging this. Thanks!

Issue #321, opened 2020-08-30
Failure to build with --with-ntl


jengelh:
Building arb-2.18.1:
```
» ./configure --with-ntl
FLINT_LIB_DIR set to /usr/local/lib
FLINT_INC_DIR set to /usr/local/include
Configuring...x86_64-Linux
Testing __builtin_popcountl...yes
Testing native popcount...yes
Testing __thread...yes
ARB was successfully configured.
» make -j16
make[2]: Entering directory '~/arb/arb_fmpz_poly'
    CC   ../build/arb_fmpz_poly/deflate.lo
    CC   ../build/arb_fmpz_poly/complex_roots.lo
    CC   ../build/arb_fmpz_poly/deflation.lo
    CC   ../build/arb_fmpz_poly/evaluate_acb.lo
    CC   ../build/arb_fmpz_poly/evaluate_acb_horner.lo
    CC   ../build/arb_fmpz_poly/evaluate_acb_rectangular.lo
    CC   ../build/arb_fmpz_poly/evaluate_arb_horner.lo
    CC   ../build/arb_fmpz_poly/evaluate_arb.lo
    CC   ../build/arb_fmpz_poly/evaluate_arb_rectangular.lo
    CC   ../build/arb_fmpz_poly/gauss_period_minpoly.lo
    CC   ../build/arb_fmpz_poly/inlines.lo
    CC   ../build/arb_fmpz_poly/../arb_fmpz_poly.lo
make[2]: Leaving directory '~/arb/arb_fmpz_poly'
make[2]: Entering directory '~/arb'
make[2]: *** No rule to make target 'build/interfaces/NTL-interface.lo'.  Stop.
make[2]: Leaving directory '~/arb'
g++: error: build/interfaces/NTL-interface.lo: No such file or directory
make[1]: *** [Makefile:129: libarb.so.2.9.1] Error 1
make[1]: Leaving directory '~/arb'
make: *** [Makefile:158: library] Error 2
```
Given there is no NTL-interface.c to be found, I guess with-ntl should be purged from the configure script.

PR #342, opened 2020-09-25
do not try to build ntl interface

p15-git-acc:
closes https://github.com/fredrik-johansson/arb/issues/321

The arb configure and Makefile.in files are derived from flint which has an [interfaces](https://github.com/wbhart/flint2/tree/trunk/interfaces) directory containing [NTL-interace.cpp](https://github.com/wbhart/flint2/blob/trunk/interfaces/NTL-interface.cpp) and its tests and docs. But arb doesn't have those things so this PR removes all references to them in its build scrips.


fredrik-johansson:
Thanks!

PR #344, opened 2020-09-26
in platt multieval break the j loop into blocks and use dot product

p15-git-acc:
I see about 25% speedup. It includes a custom `acb_dot_arb` function that treats a length n vector of acb structs like a length 2*n vector of arb structs. I saw this pattern in acb.h so I guess it's OK.


fredrik-johansson:
Yes, that's fine!

PR #354, opened 2020-11-15
use fmpz instead of slong for the J parameter of windowed zeta zeros

p15-git-acc:
closes https://github.com/fredrik-johansson/arb/issues/349

I think the old code would overflow at large heights if slong is 32 bits.


p15-git-acc:
Closed with the intention to build on this PR to make a zeta_zeros example program.

p15-git-acc:
reopened with zeta zeros program

p15-git-acc:
I noticed Dave Platt has uploaded an arb zeta zeros program with MIT license here https://github.com/djplatt/arb_zeta

PR #358, opened 2021-01-24
Rounding direction added to arb.h documentation

hannorein:
This would have saved me digging up the source code.


fredrik-johansson:
Thanks!

PR #362, opened 2021-03-23
Fixed bug for aliased mat_mul of window matrices

David-Berghaus:
I noticed that most `mat_mul`-functions (with `mat_mul_entrywise` being an exception) do not work properly for aliased window-matrices. The reason for this is that `mat_swap` destroys the entanglement of the window matrices with their parent matrix.
This can be seen with the following code-snippet:
```
void arb_mat_mul_aliased_window_test(){
    arb_mat_t A, B, A_window;
    arb_mat_init(A, 2, 2);
    arb_mat_init(B, 2, 2);
    arb_mat_window_init(A_window, A, 0, 0, 2, 2);
    arb_mat_one(A);
    arb_mat_ones(B);

    printf("Before multiplication: \n");
    printf("A: \n");
    arb_mat_printd(A, 10);
    printf("A_window: \n");
    arb_mat_printd(A_window, 10);
    arb_mat_mul_classical(A_window, B, A_window, 53);
    printf("After multiplication: \n");
    printf("A: \n");
    arb_mat_printd(A, 10);
    printf("A_window: \n");
    arb_mat_printd(A_window, 10);

    arb_mat_window_clear(A_window);
    arb_mat_clear(A);
    arb_mat_clear(B);
}
```
Which produces:
```
Before multiplication:
A:
[1 +/- 0, 0 +/- 0]
[0 +/- 0, 1 +/- 0]
A_window:
[1 +/- 0, 0 +/- 0]
[0 +/- 0, 1 +/- 0]
After multiplication:
A:
[1 +/- 0, 0 +/- 0]
[0 +/- 0, 1 +/- 0]
A_window:
[1 +/- 0, 1 +/- 0]
[1 +/- 0, 1 +/- 0]
```
As you can see, `A` no longer holds the correct result.

My fix performs the pointer-swaps entrywise which makes sure that the corresponding parent object gets updated accordingly. We now have to do O(N^2) pointer-swaps instead of O(1) but the performance impact should be negligible.

### DIFF
--- a/acb_mat.h
+++ b/acb_mat.h
@@ -66,6 +66,16 @@ acb_mat_swap(acb_mat_t mat1, acb_mat_t mat2)
     *mat2 = t;
 }
 
+ACB_MAT_INLINE void
+acb_mat_swap_entrywise(acb_mat_t mat1, acb_mat_t mat2)
+{
+    slong i, j;
+
+    for (i = 0; i < acb_mat_nrows(mat1); i++)
+        for (j = 0; j < acb_mat_ncols(mat1); j++)
+            acb_swap(acb_mat_entry(mat2, i, j), acb_mat_entry(mat1, i, j));
+}
+
 /* Window matrices */
 
 void acb_mat_window_init(acb_mat_t window, const acb_mat_t mat, slong r1, slong c1, slong r2, slong c2);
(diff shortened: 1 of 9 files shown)


fredrik-johansson:
I agree that this won't lead to a significant performance regression, but I'd like to think some more about possible alternative solutions before merging this. Other methods are affected by this aliasing bug as well.

fredrik-johansson:
Finally merged in Flint so I will merge this as well. Thanks!

PR #366, opened 2021-05-22
add acb_elliptic_p_prime

tthsqe12:
I guess this is for when you want p' but not p. ¯\_(ツ)_/¯


fredrik-johansson:
Yeah, good idea.

PR #374, opened 2021-07-27
Implement Graeffe Transform for arb_poly and acb_poly

MGessinger:
If `P` is a polynomial (over an arbitrary ring R), then its Graeffe transform `G` is defined through the equation `G(x^2) = P(x)P(-x)` (up to an overall sign, which is determined by the degree of P).
If L//R is the splitting field of P, then it is easy to see, that G also decomposes over L, and its roots are exactly the squares of the roots of P. In particular, in some cases the roots of G might be real even though those of P were not.
Due to the symmetry of G, we see that in computing `P(x)P(-x)`, half the terms will vanish (namely precisely the odd terms). This wastes time and memory. Instead, this implementation uses the definition in [\[Diouf, Sec. 1.2\]](https://tel.archives-ouvertes.fr/tel-00151313/document), which yields a significant performance benefit for the reason mentioned above.

This polynomial transform finds application in [Graeffe's Method \[Diouf, Sec. 3\]](https://tel.archives-ouvertes.fr/tel-00151313/document) for bounding polynomial roots (or Fujiwara's bound, as is already implemented in `acb_poly_root_bound_fujiwara`).
For the application of said method, the Graeffe Transform will generally be applied many times in a row. As such, it might be of interest to provide another method that takes (alongside the usual arguments) an additional `iterations` parameter, and performs the transform that many times, as opposed to calling the usual function over and over again. The benefit of having such an additional function is to avoid excessive internal memory (de-)allocations.

Tests have been provided for `arb_poly_graeffe_transform` and `acb_poly_graeffe_transform`, using the file `t-acb_poly_borel_transform.c` as style reference. Since the code is identical for the real and complex transform, the tests are also identical. Sadly, the transform doesn't decompose into real and imaginary parts, so code duplication cannot be avoided.

### DIFF
--- a/acb_poly/graeffe_transform.c
+++ b/acb_poly/graeffe_transform.c
@@ -1,0 +1,52 @@
+/*
+    Copyright (C) 2012 Fredrik Johansson
+
+    This file is part of Arb.
+
+    Arb is free software: you can redistribute it and/or modify it under
+    the terms of the GNU Lesser General Public License (LGPL) as published
+    by the Free Software Foundation; either version 2.1 of the License, or
+    (at your option) any later version.  See <http://www.gnu.org/licenses/>.
+*/
+
+#include "acb_poly.h"
+
+void
+_acb_poly_graeffe_transform(acb_ptr b, acb_srcptr a, slong len, slong prec)
+{
+    if (len == 0)
+        return;
+
+    slong q, i;
+
+    q = (len-1)/2+1;
+    acb_ptr pe = _acb_vec_init(q);
+    acb_ptr po = _acb_vec_init(len);
+
+    for (i = len-1; i >= 0; i--)
+    {
+        if (i % 2 == 0)
+            acb_set(pe+i/2, a+i);
+        else
+            acb_set(po+i/2, a+i);
+    }
+
+    _acb_poly_mul(b, po, q, po, q, prec);
+    _acb_poly_shift_left(b, b, len-1, 1);
+    _acb_poly_mul(po, pe, q, pe, q, prec);
+    _acb_vec_sub(b, po, b, len, prec);
+
+    _acb_vec_clear(pe, q);
+    _acb_vec_clear(po, len);
+
+    if (len % 2 == 0)
+	    _acb_vec_neg(b, b, len);
+}
+
+void
+acb_poly_graeffe_transform(acb_poly_t b, const acb_poly_t a, slong prec)
+{
+    acb_poly_fit_length(b, a->length);
+    _acb_poly_graeffe_transform(b->coeffs, a->coeffs, a->length, prec);
+    _acb_poly_set_length(b, a->length);
+}
--- a/acb_poly/graeffe_transform.c
+++ b/acb_poly/graeffe_transform.c
@@ -19,20 +19,20 @@ _acb_poly_graeffe_transform(acb_ptr b, acb_srcptr a, slong len, slong prec)
 
     slong q, i;
 
-    q = (len-1)/2+1;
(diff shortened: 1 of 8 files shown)


fredrik-johansson:
Looks good, though the documentation is missing.

Also, please follow the code style of putting spaces around binary operators (x + y instead of x+y).

I think you can speed up the test code as well. There's no reason to test with up to n = 200 roots; n = 20 should be more than enough.

MGessinger:
Thank you for your feedback. I have provided documentation in what I believe to be the correct places, as well as implementing the code changes you suggested.
If you spot any further issues, I will happily see to them.

acb_poly/graeffe_transform.c
```diff
@@ -0,0 +1,52 @@
...
+
+#include "acb_poly.h"
+
+void
+_acb_poly_graeffe_transform(acb_ptr b, acb_srcptr a, slong len, slong prec)
+{
+    if (len == 0)
+        return;
+
+    slong q, i;
```
fredrik-johansson, line 20: Variable declarations should come before any code.

acb_poly/test/t-graeffe_transform.c
```diff
@@ -0,0 +1,85 @@
...
+{
+    slong iter;
+    flint_rand_t state;
+
+    flint_printf("graeffe_transform....");
+    fflush(stdout);
+
+    flint_randinit(state);
+
+    acb_poly_t a, b, c;
```
fredrik-johansson, line 24: Here as well.

acb_poly/graeffe_transform.c
```diff
@@ -0,0 +1,52 @@
...
+
+    for (i = len - 1; i >= 0; i--)
+    {
+        if (i % 2 == 0)
+            acb_set(pe + i / 2, a + i);
+        else
+            acb_set(po + i / 2, a + i);
+    }
+
+    _acb_poly_mul(b, po, q, po, q, prec);
```
fredrik-johansson, line 34: If the polynomial has even degree, then the odd part will be 1 shorter than the even part. Using this exact length will be a small optimization. You can also skip the _acb_poly_shift_left by writing directly to b + 1.

MGessinger:
The stylistic mistakes were fixed, and the suggested optimization was implemented. Both arrays (for the even and off terms) now have minimal length, and all polynomial operations use as few terms as possible.
I also used the simplification of writing to `b+1` directly. Note that while I could do this "trick" right when I perform the multiplication (line 42 in *graeffe_transform.c* currently), that would require me to set the lowest term to zero. Instead, I use this "trick" one step later, which avoids that inconvenience.

In optimizing the array lengths, one encounters a possible problem, if `b` already contains data. To solve this problem, the leading coefficient of `b` is set manually in the case that the degree is odd (i.e. the length is even). In the other case, this problem does not occur.

fredrik-johansson:
Excellent, thanks!

fredrik-johansson:
By the way, a possible issue: the test code only generates monic polynomials. You might want to test non-monic polynomials as well to make sure the leading coefficient is handled correctly.

MGessinger:
Ah, good point. I altered the test code to include a randomly generated prefactor, and indeed it turns out that the leading coefficient was not handled correctly, albeit only in the case of constant input polynomials (which, given that Graeffe's method is used for finding roots, should hardly ever be of interest, but still needs to be covered). This has been fixed now.
How would I go about having the fix merged? Do I open a new Pull Request? Apologies for the inconvenience.

fredrik-johansson:
No problem. You can just make a new PR.

PR #377, opened 2021-09-21
Add acb_add_error_arb

albinahlback:
Also add acb_add_error_arf to the documentation.

This is nice to have when developing libraries where error-bounds need to be calculated using arbs (for example if they involve special functions like upper incomplete gamma functions etc.).

Let me know if anything can be improved, or if tests are needed.


fredrik-johansson:
arb_add_error_arb should be arb_add_error

PR #383, opened 2021-10-18
Compute bin_uiui via Flint for small n

albinahlback:
Big speedup for smaller `n` and increased precision due to not using floating point precision.

Let me know if you want any timings for this.

### DIFF
--- a/arb/bin.c
+++ b/arb/bin.c
@@ -1,6 +1,6 @@
 /*
     Copyright (C) 2013 Fredrik Johansson
-    Copyright (C) 2021 Albin Ahlbäck
+    Copyright (C) 2021, 2022 Albin Ahlbäck
 
     This file is part of Arb.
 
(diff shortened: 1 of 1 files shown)


albinahlback:
Does it fail because MPIR might be used?

albinahlback:
Does this look good @fredrik-johansson ?

fredrik-johansson:
It's ok, but I'd like to see cutoffs based on n, k and the precision.

Computing bin(n, k) via GMP should be faster even for huge n, if k is small or if log2(bin(n, k)) is less than some small multiple of prec.

It would be nice if also arb_bin_ui called the same algorithm when n is a small integer.

albinahlback:
Yes, that's true! Thanks for your input.

albinahlback:
I haven't tested the code yet, but I think the thoughts are solid.

I profiled with 10 bits in order to give the original code an advantage, and found that the bounds in the code are pretty good estimates for when the new version is faster. I had to do some extra code in case that `ulong != unsigned long` and in case that `FLINT_BITS != 64`.

arb/bin.c
```diff
@@ -10,42 +11,106 @@
 */
 
 #include "arb.h"
+#include "flint/fmpz.h"
+
+#define LOG2_BIN(n, k)                      \
+    (-1.3257480647361592                    \
+     + ((n) + .5) * log2(n)                 \
```
fredrik-johansson, line 18: I think log2 might not be available everywhere. Better to use log.

arb/bin.c
```diff
@@ -10,42 +11,106 @@
...
+        mpz_t mres;
+        mpz_init(mres);
+        flint_mpz_bin_uiui(mres, n, k);
+        arb_set_round_mpz(x, mres, prec);
+        mpz_clear(mres);
+    }
+    else if (k < (UWORD(1) << 32) && ((double) prec) < LOG2_BIN(n, k))
+    {
+        mpz_t mres;
+        __mpz_struct mn = {1, 1, NULL};
```
fredrik-johansson, line 98: This level of code shouldn't be be written in a way that relies on the internal representation of structs.

arb/bin.c
albinahlback: What is your opinion about supporting/assuming newer standards?

PR #387, opened 2021-10-23
add profiling code for arb_const_pi

tthsqe12:
This is a fun exercise in pushing the pi code. It is fairly careful crafted, but it still may or may not work or be correct.

### DIFF
--- a/arb/profile/p-const_pi.c
+++ b/arb/profile/p-const_pi.c
@@ -1,5 +1,23 @@
-/* usage: make profile && ./build/arb/profile/p-const_pi */
-/* untested with FLINT_BITS != 64 !!! */
+/*
+    This is a significant rewrite of gmp-chudnovsky.c by Hanhong Xue with at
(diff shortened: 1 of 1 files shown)


fredrik-johansson:
Fabrice Bellard claims to have some ways to save memory in https://bellard.org/pi/pi2700e9/pipcrecord.pdf - I presume you already read this?

For speed, there is also the trick of precomputing a table of powers of the argument of the hypergeometric series. I've used this for a few other binary splitting functions and it gives a small speedup (not sure how significant here).

tthsqe12:
It has one fewer bug now (ui_factor_mulpow_inplace), and the majority of the code is reusable general-purpose parts now. Not sure what the point of precomputing some powers of the argument is as they are walking around factored anyways. Maybe we can use the fact that the final denominator is known...

fredrik-johansson:
You no longer want to work on this?

PR #396, opened 2021-12-21
[WIP] Split `arb_pow_fmpz_binexp` into a `ui` and `mpz` version and implement proper `arb_sqr`

albinahlback:
I am done with the former part, with addition of adding `nn_sqr_2` which simplifies `nn_mul_2x2` by skipping one `umul_ppmm`.

I am going to add `arf_sqr_rnd_down` and then implement `arb_sqr` from there.


arb/sqr.c
```diff
@@ -0,0 +1,67 @@
...
+        else
+            mag_zero(arb_radref(res));
+    }
+    else if (ARB_IS_LAGOM(x) && ARB_IS_LAGOM(res))
+    {
+        mag_fast_init_set_arf(xm, arb_midref(x));
+
+        mag_init(resr);
+        mag_fast_mul(resr, xm, arb_radref(x));
+        if (resr->exp < COEFF_MAX && resr->exp >= COEFF_MIN)
```
fredrik-johansson, line 35: These checks are not needed; the exponent can be increment with risk of overflow.

arb/sqr.c
```diff
@@ -0,0 +1,67 @@
...
+
+        *arb_radref(res) = *resr;
+    }
+    else
+    {
+        mag_init_set_arf(xm, arb_midref(x));
+
+        mag_init(resr);
+        mag_mul(resr, xm, arb_radref(x));
+        fmpz_add_ui(&(resr->exp), &(resr->exp), 1);
```
fredrik-johansson, line 54: Use MAG_EXPREF

arf.h
```diff
@@ -914,6 +926,12 @@ int arf_mul_rnd_down(arf_ptr z, arf_srcptr x, arf_srcptr y, slong prec);
         ? arf_mul_rnd_down(z, x, y, prec)        \
         : arf_mul_rnd_any(z, x, y, prec, rnd))
 
+int arf_sqr_via_mpfr(arf_ptr res, arf_srcptr x, slong prec, arf_rnd_t rnd);
```
fredrik-johansson, line 929: This function is pointless extra code; arf_mul_via_mpfr already does the job.

doc/source/arf.rst
```diff
@@ -111,6 +111,18 @@ Types, macros and constants
     test code. If in doubt, simply try with some convenient high precision
     instead of using this special value, and check that the result is exact.
 
+.. macro:: nn_mul_2x1(r2, r1, r0, a1, a0, b0)
+
+    Sets `(r_2, r_1, r_0)` to `(a_1, a_0)` multiplied by `b_0`.
```
fredrik-johansson, line 116: Might want to document that these are limbs. Also, I think these macros don't allow aliasing, which might be worth documenting.

doc/source/arf.rst
```diff
@@ -553,7 +565,8 @@ Addition and multiplication
 
 .. function:: int arf_neg_round(arf_t res, const arf_t x, slong prec, arf_rnd_t rnd)
 
-    Sets *res* to `-x`.
+    Sets *res* to `-x`. Returns 0 if this operation was made exactly and 1 if
+    truncation occurred.
```
fredrik-johansson, line 569: "truncation" is the wrong term (means rounding down) - should be "rounding".

mag.h
```diff
@@ -112,6 +112,23 @@ _fmpz_sub2_fast(fmpz_t z, const fmpz_t x, const fmpz_t y, slong c)
     }
 }
 
+static __inline__ void
+_fmpz_2times_fast(fmpz_t res, const fmpz_t x, slong c)
+{
+    slong cx = *x;
+
+    if (!COEFF_IS_MPZ(*res) && (cx > COEFF_MIN / 2 && cx < COEFF_MAX / 2))
```
fredrik-johansson, line 120: These conditions are wrong. See _fmpz_add2_fast. Anyway, I think this function is not really needed; the compiler should be able to eliminate the redundant branches in _fmpz_add2_fast when passed the same operand twice.

fredrik-johansson:
Did you do some profiling?

mag.h
albinahlback: It has actually one property that I do not think the compiler can transform into, and that is that we now can allow `cx` to be up to size `COEFF_MAX / 2` instead of `COEFF_MAX / 4`.

arb/sqr.c
albinahlback: I presume you meant _without_  risk of overflow ;) Och god jul, Fredrik!

mag.h
fredrik-johansson: `_fmpz_add2_fast`/`_fmpz_sub2_fast` need to work for values of `c` larger than +/- 1. I think it's confusing if `_fmpz_2times_fast` in contrast to the other functions only supports `|c| <= 1`.

In practice, it makes no difference to go up to `COEFF_MAX / 2` instead of `COEFF_MAX / 4` since such huge exponents are very rare. We only care about optimizing for small exponents here. Better to be uniform and conservative.

Considering this, I don't really think this extra function is needed. 

Clearly these functions and their assumptions should be documented...

albinahlback: Okay, will remove it then.

arf.h
albinahlback: It removes one branching and removes two/three useless assignments/calculations. My thinking with this PR is to strip all the "unnecessary" stuff away from squaring in order to make it as fast as possible. I know it's pretty small improvement, but you do not think that it's worth it?

fredrik-johansson: mul_via_mpfr is only used for products that take 1000+ cycles to compute so a few cycles here will not be measurable.

arb/sqr.c
fredrik-johansson: God jul :)

Issue #397, opened 2021-12-21
cl, ARCH = 64 fails due to AppVeyor not finding GMP


albinahlback:
See the new PR's.

PR #398, opened 2021-12-21
update conda location in appveyor

isuruf:
Fixes https://github.com/fredrik-johansson/arb/issues/397


fredrik-johansson:
Thanks!

Issue #408, opened 2022-03-10
Should positive * pos_inf be pos_inf instead of zero_pm_inf?


postmath:
I've seen your comments that you feel that the various non-number values in arb are not well enough defined, but I think this should be a case where there is a reasonable mathematically correct interpretation.

In `arb_mul` (and `arb_mul_arf`), if you multiply an `arb_t` (or an `arf_t`) representing a signed infinity by a non-exact finite `arb_t`, then the result's radius is infinite. It seems like it would be mathematically sound to test whether the finite `arb_t` is either all positive or all negative, and if so, set the result's radius to be finite instead -- correct? That would give answers that users might expect more, even though of course the current result is also correct. I would be happy to prepare a patch.

fredrik-johansson:
Yes, this seems reasonable.

PR #409, opened 2022-03-12
Positive * pos_inf should be pos_inf, not 0 +- inf. Same for other combinations of signs.

postmath:
This addresses github issue #408.


fredrik-johansson:
Thanks! Just two changes:
* Move these branches below ARB_IS_LAGOM(...). These are fast paths for regular arithmetic and exclude infinities.
* Add some test code (does not need to be randomized tests; you can just test the relevant special cases).

postmath:
Thanks for the review.

fredrik-johansson:
Looks great. Thanks!

PR #418, opened 2022-04-13
Fixing MINGW64 build

mazar:
ARB does not build correctly in `MINGW64`. `-std=c99` needs to be added to `CFLAGS` in `configure` script. Also copied some more logic related to auto detection of `OS` and `ARCH` from https://github.com/fredrik-johansson/flint2/blob/trunk/configure

### DIFF
--- a/configure
+++ b/configure
@@ -317,11 +317,23 @@ KERNEL=`uname`
 if [ -z "$BUILD" ]; then
    ARCH=`uname -m`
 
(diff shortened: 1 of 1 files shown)


fredrik-johansson:
Thanks. Since the code is the same as in the flint configure (as far as I can tell) there shouldn't be any issue with it, assuming that you've tested it. Will merge immediately.

mazar:
@fredrik-johansson any plans on releasing a new tag with the latest commits?

PR #424, opened 2022-06-01
add lerch_phi to fpwrap

vale981:
This is only a proposal. I haven't tested this.

### DIFF
--- a/arb_fpwrap/fpwrap.c
+++ b/arb_fpwrap/fpwrap.c
@@ -981,6 +981,28 @@ _arb_polygamma(arb_t res, const arb_t s, const arb_t z, slong prec)
 }
 
 static void
+_arb_dirichlet_lerch_phi(arb_t res, const arb_t z, const arb_t s, const arb_t a, slong prec)
+{
+    acb_t t, u, v, w;
+    acb_init(t);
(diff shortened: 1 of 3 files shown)


vale981:
done :). after the CI is through, i'll remove the accidental whitespace change

vale981:
The tests work in the CI but I can't get it to compile locally.
Some outdated depency I think: ``bsplit_basecase_func_t`` is undeclared.

fredrik-johansson:
> Some outdated dependency I think: bsplit_basecase_func_t is undeclared.

What version of flint are you building against? It should be present in the current git version (wbhart's repo) and the 2.9 beta.

fredrik-johansson:
I forgot about docs: could you just add the functions to `doc/source/arb_fpwrap.rst` as well? Should be ready to merge after that!

vale981:
done that :)

fredrik-johansson:
> you're the most responsive repo-owner i've encountered so far

Thank you. 

(I'm not always responsive 😬 You were lucky to submit this on a good day...)

vale981:
thank you for this nice project. it's hard to find a fast-ish lerch transcendent otherwise :)

PR #435, opened 2022-07-11
add fmpzi_gcd_shortest

tthsqe12:
This is the nemo version


tthsqe12:
Sure, but unfortunately its complexity has completely different characteristcs from the Euclidean algorithm. If Euclid does just a few iterations, it will be hard to beat that.

fredrik-johansson:
So for huge numbers, you'd want a version of the Euclidean algorithm that quits after a few iterations?

tthsqe12:
Ideally you would want a quasi-linear algorithm that deals directly with the quotient sequence, but that is a lot of work, so this is a compromise.

PR #443, opened 2022-11-11
Issue 441: create arb_nonnegative_abs.

postmath:
Note the following behaviour. Taking the "nonnegative abs" of `[0 \pm 1]`:
```
arb_t t;
arb_init(t);
arf_zero(arb_midref(t));
mag_one(arb_radref(t));

arb_nonnegative_abs(t, t);
```
you don't get the exact interval `[0.5 \pm 0.5]`, but instead the centre and radius are one ulp (of a `mag_t`) greater than that. That ulp is introduced in `arf_get_mag`. I'm not sure there is an obvious workaround for that. (The example I ran into started with `[0 \pm 0.25]`, but I think `[0 \pm 1]` should work the same.)

Note furthermore that the "whole real line" object can be returned by this function, which still contains negative numbers. So the result does *not* always satisfy `arb_is_nonnegative`. Neither does NaN, of course. I think those are the only two cases.

### DIFF
--- a/arb.h
+++ b/arb.h
@@ -165,6 +165,7 @@ void arb_neg(arb_t y, const arb_t x);
 void arb_neg_round(arb_t x, const arb_t y, slong prec);
 
 void arb_abs(arb_t y, const arb_t x);
+void arb_nonnegative_abs(arb_t y, const arb_t x);
 
 void arb_sgn(arb_t res, const arb_t x);
 int arb_sgn_nonzero(const arb_t x);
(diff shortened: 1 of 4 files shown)


fredrik-johansson:
Looks good. If you prefer a different name for this function, I don't mind changing it.

`arf_get_mag` rounds inexactly for speed reasons. It would indeed be better to make it round exactly, so that cases like this work as expected, having an `arf_get_mag_sloppy` for the cases where saving the 1-2 cycles matters.

postmath:
No need for the other name, I changed my mind and I like this version now :)

